PR #74, opened 2023-10-16
feat(rsbuild/shared): init @rsbuild/shared

9aoy:
## Checklist


packages/shared/compiled/autoprefixer/index.d.ts
```diff
@@ -0,0 +1 @@
+export = any;
```
chenjiahan, line 1: Should add "compiled" folder to packageJson.files

pnpm-lock.yaml
```diff
@@ -139,26 +139,166 @@ importers:
...
+        specifier: 5.19.2
+        version: 5.19.2
+      terser-webpack-plugin:
+        specifier: 5.3.9
+        version: 5.3.9(webpack@5.89.0)
       typescript:
         specifier: ^5
         version: 5.2.2
 
+  packages/shared/compiled/autoprefixer: {}
```
chenjiahan, line 228: Can we skip compiled package.json in our pnpm workspace config.

PR #111, opened 2023-10-18
refactor(builder): extract some builder methods

9aoy:
## Checklist


packages/core/package.json
```diff
@@ -33,6 +33,10 @@
       "types": "./dist/plugins/*.d.ts",
       "default": "./dist/plugins/*.js"
     },
+    "./base/*": {
```
chenjiahan, line 36: After extracting these codes, it seems that the number of lines of code and complexity are even higher.

image

9aoy: Type calculation is more complex, subsequent processing 😢

chenjiahan:
ref: https://github.com/web-infra-dev/rsbuild/discussions/239

Issue #73, opened 2023-10-16
Adjust the default browserslist config


chenjiahan:
## Overall

Rsbuild will by default target modern browsers to reduce bundle sizes.

## Detail

The default browserslist config of Modern.js Builder:

```yaml
> 0.01%
not dead
not op_mini all
```

The default browserslist config of Rsbuild:

```js
const browserslist = [
  'chrome >= 87',
  'edge >= 88',
  'firefox >= 78',
  'safari >= 14',
];
```

This change should not affect Modern.js Framework.

PR #114, opened 2023-10-19
feat: ship less polyfills by default

chenjiahan:
## Related Issue

https://github.com/web-infra-dev/rsbuild/issues/73

## Checklist


packages/document/docs/en/guide/advanced/browserslist.md
```diff
@@ -118,11 +118,10 @@ If you don't need to be compatible with IE 11 browsers, you can adjust Browsersl
 - Set to browsers that supports native ES Modules (recommended):
```
s-cork, line 118: @chenjiahan i was trying to work out where the default browserslist was derived from
and it looks like this line was never updated with this PR

I think it should read something else after this update

chenjiahan: Could you create an issue to explain your requirements?

PR #341, opened 2023-11-03
fix(shared): get ipv4 interfaces should be uniq & set @types/fs-extra…

southorange0929:
… as devDeps

## Summary

Sometimes, we will get many ipv4 address.But some are same,so address should be uniq.
image

## Checklist

### DIFF
--- a/packages/shared/src/url.ts
+++ b/packages/shared/src/url.ts
@@ -1,6 +1,5 @@
 import os from 'os';
 import { URL } from 'url';
-import { uniq } from 'lodash';
 import urlJoin from 'url-join';
 import { DEFAULT_DEV_HOST } from './constants';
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/url.ts
```diff
@@ -37,7 +38,7 @@ const getIpv4Interfaces = () => {
       }
     });
   });
-  return ipv4Interfaces;
+  return uniq(ipv4Interfaces);
```
chenjiahan, line 41: Thanks for your contribution! 😄

I'm wondering that if `lodash.uniq` can be used to uniq objects.

I have written some simple test and it does not work:

image

Also Rsbuild will remove `lodash` from dependencies in the future (to reduce dependencies)

So can we use another method to uniq the ipv4Interfaces?

southorange0929: Got it,let me see.

southorange0929: I use map to uniq address, i'm not sure map is good choose for us.But i think it's better than object

chenjiahan (approved):

chenjiahan:
LGTM!

Issue #3, opened 2023-10-09
Redesign the dev server, decouple with Modern.js server


chenjiahan:
Rsbuild should avoid depending on Modern.js packages, so we need to design a new dev server for Rsbuild, and allow Modern.js to replace the builtin dev server.

9aoy:
The goal of Rsbuild Server is to be lightweight, simple, and easy to expand:
- Provide dev/preview capability for rsbuild application
- Meet the general configuration capabilities used when starting an application (such as setMiddleware, hmrClient, historyApiFallback,etc).
- Support the expansion of server capabilities to meet the complex needs of upper-level frameworks (such as modern.js) for servers

### Detail Design

Rsbuild Server uses a minimalist combination of `http + connect` , with all built-in/extended logic added via middleware.

```
var connect = require('connect');
var http = require('http');
 
var app = connect();

app.use(function (req, res, next) {
  // req is the Node.js http request object
  // res is the Node.js http response object
  // next is a function to call to invoke the next middleware
})

//create node.js http server and listen on port
http.createServer(app).listen(3000);
```

#### devServers comparison:
|                              | Rsbuild DevServer | Modern.js DevServer (old)                        | webpack-dev-server |
| ---------------------------- | ----------------- | ------------------------------------------------ | ------------------ |
| Middleware mode              | connect           | modern.js custom                                 | express            |
| Relationship with prodServer | Shared Middleware | DevServer inherits prodServer.                   | No prodServer      |
| depends on route.json        | ❌                 | ✔️ (depends on route.json and routing Middleware) | ❌                  |
| support custom middlewares   | ✔️                 | ✔️                                                | ✔️                  |
| support custom plugins       | ❌                 | ✔️                                                | ❌                  |
| support static directories   | ❌                 | ✔️                                                | ✔️                  |
| ssr support                  | ❌                 | ✔️                                                | ❌                  |
| proxy support                | ✔️                 | ✔️                                                | ✔️                  |


#### Breaking changes

Compared to modern.js builder, breaking changes include:
- Only match file based on static file paths. For example, if you want to access the page through `/`, you need to ensure that the html output path is `dist/index.html`
- `tools.devServer.before/after` configuration is not supported, you can use `tools.devServer.setupMiddlewares` instead

PR #405, opened 2023-11-08
feat: init devServer

9aoy:
## Summary
Init rsbuild devServer, decouple with Modern.js server.

Compared to modern.js builder, breaking changes include:
- Only match file based on static file paths. For example, if you want to access the page through /, you need to ensure that the html output path is dist/index.html
- SPA and multi-entry SPA are not supported by default, this can be achieved by configuring tools.devServer.historyApiFallback
- tools.devServer.before/after configuration is not supported, you can use tools.devServer.setupMiddlewares instead

TODO:
- support rsbuild prodServer
- add more middlewares
- remove all Modern.js server deps in rsbuild

## Related Issue

https://github.com/web-infra-dev/rsbuild/issues/3

## Checklist


packages/core/src/server/devServer.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    this.middlewares.use((req, res, next) => {
+      // allow hmr request cross-domain, because the user may use global proxy
+      res.setHeader('Access-Control-Allow-Origin', '*');
+      const path = req.url ? url.parse(req.url).pathname : '';
+      if (path?.includes('hot-update')) {
+        res.setHeader('Access-Control-Allow-Credentials', 'false');
+      }
+
+      // 用户在 devServer 上配置的 headers 不会对 html 的请求生效，加入下面代码，使配置的 headers 对所有请求生效
```
chenjiahan, line 104: English by default

chenjiahan, line 104:
```suggestion
      // The headers configured by the user on devServer will not take effect on html requests. Add the following code to make the configured headers take effect on all requests.
```

chenjiahan (approved):

Issue #448, opened 2023-11-10
[Bug]: entry has to be accessed with the suffix `.html` and didn't logged out in the terminal


zoolsher:
### Version

```sh
System:
    OS: macOS 14.0
    CPU: (10) arm64 Apple M1 Pro
    Memory: 68.09 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 119.0.6045.123
    Edge: 119.0.2151.58
    Safari: 17.0
  npmPackages:
    @rsbuild/core: ^0.0.19 => 0.0.19 
    @rsbuild/plugin-react: ^0.0.19 => 0.0.19
```

### Details

created a project for my school assignment and uploaded it to [Github](https://github.com/zoolsher/rsbuild-repro-entry). I made an entry on rsbuild following the guidelines provided on its official website as shown below:

```ts
//...
{
   "foo": "./src/foo.tsx",
}
//...
```

After starting the development server, the following was displayed on the terminal:

```shell
Rsbuild v0.0.19

info    Port 8080 is in use, using port 8081.

  > Local:    http://localhost:8081
  > Network:  http://10.3.213.5:8081
  > Network:  http://10.254.243.239:8081

start   Compiling...
ready   Client compiled in 0.25 s
```

When I clicked on the link provided in the terminal, it launched localhost:8081. However, when I attempted to access localhost:8081/foo, it returned a 404 error.

----

My suggestion would be

1. print the links that can be accessed in the terminal like
```
Rsbuild v0.0.19

info    Port 8080 is in use, using port 8081.

  > Local:    http://localhost:8081
  > Network:  http://10.3.213.5:8081
  > Network:  http://10.254.243.239:8081

Entry list:

foo: http://localhost:8081/foo.html
index: http://localhost:8081/index.html

start   Compiling...
ready   Client compiled in 0.25 s

```

2. follow the config provided by developers and only add the suffix when they are doing something like this:

```ts
//...
{
  "foo.html": "./src/pages/foo.tsx"
}
//...
```

### Reproduce link

https://github.com/zoolsher/rsbuild-repro-entry

### Reproduce Steps

1. pnpm install
2. open `https://localhost:8080/foo`

### additional information

Vite uses an HTML file as the entry point, but it also supports access without the `.html` suffix.

![image](https://github.com/web-infra-dev/rsbuild/assets/9161085/6672d976-be1e-4367-84eb-b68a30f1e70f)

[...]

PR #468, opened 2023-11-11
feat(server): should print correct urls by html entry

9aoy:
## Summary

entry (single): index
image

entry (single): foo
image

entry(multiple): foo、bar
image

## Related Issue

fix: #448 

## Checklist

### DIFF
--- a/packages/core/src/plugins/startUrl.ts
+++ b/packages/core/src/plugins/startUrl.ts
@@ -4,7 +4,7 @@ import {
   castArray,
   type DefaultRsbuildPlugin,
   normalizeUrl,
-  getRoutesName,
+  formatRoutes,
 } from '@rsbuild/shared';
 import { exec } from 'child_process';
 import { promisify } from 'util';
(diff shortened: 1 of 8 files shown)


packages/core/src/plugins/startUrl.ts
```diff
@@ -102,8 +108,13 @@ export function pluginStartUrl(): DefaultRsbuildPlugin {
         const urls: string[] = [];
 
         if (startUrl === true || !startUrl) {
+          const { entry } = api.context;
+          const names = getRoutesName(entry);
           const protocol = https ? 'https' : 'http';
-          urls.push(`${protocol}://localhost:${port}`);
+          // auto open first entry
+          urls.push(
+            normalizeUrl(`${protocol}://localhost:${port}/${names[0]}`),
```
chenjiahan, line 116: If one of the names is `index`, can we just open `http://localhost:8080` instead of `http://localhost:8080/index` or `http://localhost:8080/xxx` ?

packages/shared/src/devServer.ts
```diff
@@ -14,13 +15,40 @@ import { DEFAULT_PORT, DEFAULT_DEV_HOST } from './constants';
 import { createAsyncHook } from './createHook';
 import type { Compiler } from '@rspack/core';
 
+export const normalizeUrl = (url: string) => url.replace(/([^:]\/)\/+/g, '$1');
```
chenjiahan, line 18: Can we add some unit tests for this method, RegExp is difficult to understand.

packages/core/src/plugins/startUrl.ts
9aoy: done

packages/shared/src/devServer.ts
9aoy: done

chenjiahan (approved):

chenjiahan:
👍🏻 Good job

PR #530, opened 2023-11-15
refactor: always use source.entries to set entry

chenjiahan:
## Summary

Always use `source.entries` to set entry, remove the `entry` option of `createRsbuild`.

```ts
// before
createRsbuild({
  entry: {}
});

// after
createRsbuild({
  rsbuildConfig: {
    source: {
      entries: {}
    }
  }
});
```

## Checklist

- [x] I have added changeset via `pnpm run change`.
- [x] I have updated the documentation.


chenjiahan:
!canary

PR #600, opened 2023-11-20
fix(devServer): should render html success when set output.publicPath

9aoy:
## Summary

Align with webpack-dev-middleware, get publicPath from compiler. 
Previously, rsbuild get assetPrefix as publicPath, but some users would set output.publicPath in modifyingRspackConfig, so assetPrefix was not accurate.

## Checklist


packages/core/src/server/devServer.ts
```diff
@@ -249,19 +251,22 @@ export async function startDevServer<
...
     devMiddleware,
     dev: devServerConfig,
     output: {
       distPath: rsbuildConfig.output?.distPath?.root || ROOT_DIST_DIR,
-      assetPrefix:
-        typeof devServerConfig.assetPrefix === 'string' &&
-        !isURL(devServerConfig.assetPrefix)
-          ? devServerConfig.assetPrefix
-          : '',
+      publicPaths,
```
chenjiahan, line 269: Can we just pass the first publicPath?

chenjiahan (approved):

PR #650, opened 2023-11-22
fix(core): printFileSize throw err when set output.filename query

fireairforce:
## Checklist


packages/core/src/plugins/fileSize.ts
```diff
@@ -49,14 +49,15 @@ async function printFileSizes(stats: Stats | MultiStats, distPath: string) {
   const { default: gzipSize } = await import('@rsbuild/shared/gzip-size');
 
   const formatAsset = (asset: StatsAsset) => {
-    const contents = fse.readFileSync(path.join(distPath, asset.name));
+    const fileName = asset.name;
```
chenjiahan, line 52:
```suggestion
    const fileName = asset.name.split('?')[0];
```

chenjiahan (approved):

chenjiahan:
Awesome!

PR #662, opened 2023-11-23
fix: compatible loader's rule.use maybe empty array

easy1090:
## Summary
Compatible loader's rule.use maybe empty array.

## Related Links

#663

## Checklist

- [x] I have added changeset via `pnpm run change`.
- [x] I have added tests to cover my changes.

### DIFF
--- a/packages/doctor-core/src/build-utils/build/utils/loader.ts
+++ b/packages/doctor-core/src/build-utils/build/utils/loader.ts
@@ -88,7 +88,7 @@ export function mapEachRules<T extends Plugin.BuildRuleSetRule>(
       } as unknown as T;
     }
 
-    if (rule.use) {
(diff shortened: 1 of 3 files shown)


packages/doctor-core/src/build-utils/build/utils/loader.ts
```diff
@@ -88,7 +88,11 @@ export function mapEachRules<T extends Plugin.BuildRuleSetRule>(
       } as unknown as T;
     }
 
-    if (rule.use) {
+    if (
+      rule.use &&
+      ((Array.isArray(rule.use) && rule.use.length !== 0) ||
+        !Array.isArray(rule.use))
```
chenjiahan, line 94: This condition can be simplified

```js

if (rule.use && (!Array.isArray(rule.use) || rule.use.length !== 0)) {
  //
}

```

chenjiahan (approved):

PR #701, opened 2023-11-24
feat: fix issue of service worker & add example

Asuka109:
## Summary

https://github.com/web-infra-dev/rsbuild/blob/9c1919845979029fa44d880a13dfe797825e37dd/packages/shared/src/apply/output.ts#L57-L60

This behaviour seems weird because nor service worker / web worker / edge worker have exports.

@GiveMe-A-Name provide some information:
- The service worker target currently used for the edge worker service of corp.
- It even not a complete worker destination but a module and will be warpped.

So we need to remove the `libraryTarget` option here.

Maybe should refactor target plugin of @modern-js/app-tools later.

## Checklist

### DIFF
--- a/packages/shared/src/apply/output.ts
+++ b/packages/shared/src/apply/output.ts
@@ -54,10 +54,7 @@ export function applyOutputPlugin(api: SharedRsbuildPluginAPI) {
         const workerPath = getDistPath(config.output, 'worker');
         const filename = posix.join(workerPath, `[name].js`);
 
-        chain.output
-          .filename(filename)
-          .chunkFilename(filename)
-          .libraryTarget('commonjs2');
+        chain.output.filename(filename).chunkFilename(filename);
       }
     },
   );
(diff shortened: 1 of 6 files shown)


packages/shared/src/apply/output.ts
```diff
@@ -54,10 +54,7 @@ export function applyOutputPlugin(api: SharedRsbuildPluginAPI) {
         const workerPath = getDistPath(config.output, 'worker');
         const filename = posix.join(workerPath, `[name].js`);
 
-        chain.output
-          .filename(filename)
-          .chunkFilename(filename)
-          .libraryTarget('commonjs2');
```
chenjiahan, line 60: This option is added to support Modern.js (EdenX) worker.

If we remove this, we should add this `libraryTarget` config to the Modern.js app-tools.

chenjiahan:
@Asuka109 Don't forget to add this libraryTarget config to the Modern.js app-tools.

PR #740, opened 2023-11-28
feat: support solid-js hmr

inottn:
## Summary

feat: support solid-js hmr

## Checklist

### DIFF
--- a/packages/plugin-solid/src/rename-import.ts
+++ b/packages/plugin-solid/src/rename-import.ts
@@ -1,0 +1,19 @@
+import { createVirtualModule } from '@rsbuild/shared';
+import { readFileSync } from 'fs';
+import type { PluginObj } from '@babel/core';
+
+export default function renameImport(): PluginObj {
+  return {
+    visitor: {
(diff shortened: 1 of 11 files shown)


packages/plugin-solid/src/rename-import.ts
```diff
@@ -0,0 +1,19 @@
...
+import { readFileSync } from 'fs';
+import type { PluginObj } from '@babel/core';
+
+export default function renameImport(): PluginObj {
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        if (path.node.source.value === 'solid-refresh') {
+          const runtimeFilePath = require.resolve(
+            'solid-refresh/dist/solid-refresh.mjs',
```
chenjiahan, line 11: Can we use alias to redirect the import path? The performance of alias is better than Babel plugin.

inottn: I'll give it a try~

chenjiahan:
LGTM!

PR #871, opened 2023-12-07
breaking change: add output.targets to replace createRsbuild.target

chenjiahan:
## Summary

Add output.targets to replace createRsbuild.target.

We will move the `target` option of `createRsbuild` to rsbuild config, this change allows user to configure `targets` in the rsbuild config file.

- before:

```js
const rsbuild = await createRsbuild({
  target: ['web', 'node'],
});
```

- after:

```js
// rsbuild.config.ts
export default {
  output: {
    targets: ['web', 'node'],
  }
}
```

> Only affect JavaScript API. Users using the Rsbuild CLI do not need to change.

## Related Links

https://github.com/web-infra-dev/rsbuild/discussions/813

## Checklist

- [x] Tests updated.
- [ ] Documentation updated.


chenjiahan:
!canary

PR #902, opened 2023-12-10
feat: support rsbuild.createDevServer

9aoy:
## Summary

rsbuild.createDevServer provides rsbuild server api and default middleware,  it is designed for high-level frameworks that require a custom server, such as modern.js.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


packages/shared/src/types/server.ts
```diff
@@ -85,3 +87,29 @@ export type StartServerResult = {
   port: number;
   server: ServerApi;
 };
+
+export type DevServerApi = {
```
chenjiahan, line 91:
```suggestion
export type DevServerAPI = {
```

PR #913, opened 2023-12-11
feat(server): expose devMiddlewareHooks

9aoy:
#### usage
```ts
  const devMiddlewares = await rsbuildServer.getMiddlewares({
    dev: devServerConfig,
    app: httpServer,
  });

   devMiddlewares.hooks.onChange((stats) => {
      // Reset when client compile done
      if (stats.toJson({ all: false }).name !== 'server') {
         // do something...
      }
    });
```
## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/core/src/server/compiler-dev-middleware/index.ts
+++ b/packages/core/src/server/compiler-dev-middleware/index.ts
@@ -5,6 +5,7 @@ import type {
   DevMiddlewareAPI,
   NextFunction,
   DevMiddleware as CustomDevMiddleware,
+  DevMiddlewareHooks,
 } from '@rsbuild/shared';
 import SocketServer from './socketServer';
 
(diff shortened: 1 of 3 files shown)


packages/core/src/server/compiler-dev-middleware/index.ts
```diff
@@ -46,15 +51,21 @@ export default class DevMiddleware extends EventEmitter {
...
 
     // init socket server
     this.socketServer = new SocketServer(dev);
 
     this.devMiddleware = devMiddleware;
+
+    this.eventEmitter = new EventEmitter();
+
+    this.hooks = {
+      onChange: (fn) => {
```
chenjiahan, line 65: Can we just use `rsbuild.onDevCompileDone`

Issue #825, opened 2023-12-04
[Feature]: Improve performance of PostCSS load config


chenjiahan:
### What problem does this feature solve?

According to `postcss-loader` document:

> For large projects, to optimize performance of the loader, it is better to provide postcssOptions in loader config and specify config: false. This approach removes the need to lookup and load external config files multiple times during compilation.

https://github.com/webpack-contrib/postcss-loader#postcssoptions

### What does the proposed API look like?

We can use https://www.npmjs.com/package/postcss-load-config to load PostCSS config once, and no need to load external config files multiple times.

PR #947, opened 2023-12-14
perf: improve performance of PostCSS load config

chenjiahan:
## Summary

Check out https://github.com/web-infra-dev/rsbuild/issues/825 for more information.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


inottn:
I would like to know in what scenarios `postcss-load-config` would be used instead of `postcss-loader` to load external config in the current design.

I've noticed that when a user modifies the config through `tools.postcss`, unless `config: false` is explicitly specified, `postcss-loader` will still be used to load external config. Is this behavior intentional?

chenjiahan:
In the latest Rsbuild version, `config: false` is set by default, so postcss-loader will not load `postcss.config.*` file.

Rsbuild will read `postcss.config.*` via `postcss-load-config` package, it only execute once, so the performance is better than postcss-loader's default behavior.

chenjiahan:
I understand your scenario, we need to ensure `config: false` is set even user uses `tools.postcss`.

chenjiahan:
So `config: false` should be set after we merge the postcss configs. https://github.com/web-infra-dev/rsbuild/blob/main/packages/shared/src/css.ts#L136-L140

PR welcome~

inottn:
> I understand your scenario, we need to ensure `config: false` is set even user uses `tools.postcss`.

Yes, when a user modifies the config through `tools.postcss`, the default config might be overridden. `postcss-loader` will treat `config: undefined` as `config: true`.

chenjiahan:
Yeah, I forgot to consider this scenario 😂

inottn:
There's another issue: even if the user specifies `config: true`, do we still use `postcss-load-config` to load external config? Because `postcss-load-config` is not fully compatible with `postcss-loader`, for instance, when the external config exports a function, it may receive different parameters. `postcss-loader` will pass in the loaderContext as an argument. Although, I guess this scenario might be less common. In my opinion, consistency is very important, meaning that in any situation, `postcss-load-config` should be used to load external config.

chenjiahan:
Agree, we will not support `config: true` in any scenarios, and always use `postcss-load-config`.

The function usage is supported by `tools.postcss` (but without loaderContext). Using loaderContext should be a rare scenario.

PR #968, opened 2023-12-15
test(e2e): fix some cases in Windows

chenjiahan:
## Summary

Some cases are using posix path to match the output files, so we need to adjust some conditions.

## Checklist

- [x] Tests updated.
- [ ] Documentation updated.


9aoy:
👏

PR #969, opened 2023-12-15
fix: always use postcss-load-config to load external config

inottn:
## Summary
always use `postcss-load-config` to load external config

## Related Links

relate #947 

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan (approved):

chenjiahan:
👍🏻👍🏻

PR #979, opened 2023-12-17
feat(create-rsbuild): set the name of package.json

zhousg:
## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/create-rsbuild/src/index.ts
+++ b/packages/create-rsbuild/src/index.ts
@@ -84,7 +84,7 @@ async function main() {
   const distFolder = path.join(cwd, targetDir);
 
   copyFolder(commonFolder, distFolder, version);
-  copyFolder(srcFolder, distFolder, version, targetDir);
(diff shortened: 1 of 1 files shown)


packages/create-rsbuild/src/index.ts
```diff
@@ -84,7 +84,7 @@ async function main() {
   const distFolder = path.join(cwd, targetDir);
 
   copyFolder(commonFolder, distFolder, version);
-  copyFolder(srcFolder, distFolder, version);
+  copyFolder(srcFolder, distFolder, version, targetDir);
```
chenjiahan, line 87: A folder may have multi-level paths, we can use the basename of the path instead.

image

zhousg: Get ~

chenjiahan (approved):

chenjiahan:
🎉🎉

PR #985, opened 2023-12-18
feat: support get target in tools.postcss

9aoy:
## Summary

sometimes, user need to judge the build environment by target when they want to add plugins.
 
## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan:
Only `web` target provides CSS outputs, so we may not need to get `target` in the `tools.postcss` config.

PR #993, opened 2023-12-18
fix: use rspack config chain

bvanjoi:
Using the webpack type in `﻿webpack-chain` is not an optimal choice for bundlers based on rspack.

### DIFF
--- a/packages/shared/compiled/webpack-chain/types/index.d.ts
+++ b/packages/shared/compiled/webpack-chain/types/index.d.ts
@@ -1,10 +1,19 @@
 import {
   Configuration,
   Compiler,
-  WebpackPluginInstance,
   RuleSetRule,
-  ResolveOptions,
-} from 'webpack';
+  ResolveOptions
+} from '@rspack/core';
+
+declare interface WebpackPluginInstance {
+	[index: string]: any;
+
+	/**
+	 * The run point of the plugin, required method.
+	 */
+	apply: (compiler: Compiler) => void;
+}
+
 import * as https from 'https';
 
 export = Config;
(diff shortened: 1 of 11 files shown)


bvanjoi:
related to https://github.com/web-infra-dev/rspack/issues/5045

packages/compat/webpack/tests/plugins/__snapshots__/default.test.ts.snap
```diff
@@ -2,16 +2,7 @@
...
-  "cache": {
-    "buildDependencies": {
-      "packageJson": [
-        "<ROOT>/package.json",
-      ],
-    },
-    "cacheDirectory": "<ROOT>/node_modules/.cache/webpack",
-    "name": "web-development",
-    "type": "filesystem",
-  },
```
9aoy, line 14: snapshot need update

packages/shared/src/chain.ts
```diff
@@ -257,7 +257,7 @@ export function applyScriptCondition({
 }) {
   // compile all folders in app directory, exclude node_modules
   rule.include.add({
-    and: [context.rootPath, { not: NODE_MODULES_REGEX }],
```
9aoy, line 260: Webpack supports the RegExp type, and rspack also previously supported the RegExp type. Isn't it supported now?
image

bvanjoi: see https://github.com/web-infra-dev/rspack/issues/5057

9aoy: we can revert this change~

9aoy (approved):

packages/shared/compiled/webpack-chain/types/index.d.ts
```diff
@@ -1,10 +1,19 @@
...
   Configuration,
   Compiler,
-  WebpackPluginInstance,
   RuleSetRule,
-  ResolveOptions,
-} from 'webpack';
+  ResolveOptions
+} from '@rspack/core';
+
+declare interface WebpackPluginInstance {
```
chenjiahan, line 8: We should not modify the compiled types directly, it is generated by `ncc`.

PR #1007, opened 2023-12-19
fix(plugin-html): `inject: false` should behave as expected

### LINKED ISSUE #987: [Bug]: html.inject false not works
### Version

```sh
rsbuild 0.2.5
```

### Details

 I set inject false and use custom template. expect script only inject once, but actually inject twice

rsbuildConfig: 
```ts
    rsbuildConfig: {
      html: {
        inject: false,
        template: './static/index.html',
      },
    },
```

template:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>custom template</title>
    <%= htmlWebpackPlugin.tags.headTags %>
    <script src="//example.com/assets/a.js"></script>
  </head>
  <body>
    <div id="test-template">xxx</div>
    <div id="root" />
    <%= htmlWebpackPlugin.tags.bodyTags %>
  </body>
</html>
```

### Reproduce link

https://github.com/web-infra-dev/rsbuild/blob/0759599e6d280b72a4a7bbc293be4c15b9a5316c/e2e/cases/html/inject/index.test.ts

### Reproduce Steps

- cd e2e
- pnpm test:rspack html/inject
- view dist/index.html content

xc2:
## Summary

when `html.inject` is `false`, rsbuild must not automatically inject <script> of chunks into html.

## Related Links

fixes #987

## Checklist

- [x] Tests updated.
- [x] Documentation updated: no need

### DIFF
--- a/packages/shared/src/mergeChainedOptions.ts
+++ b/packages/shared/src/mergeChainedOptions.ts
@@ -7,43 +7,74 @@ import { isFunction, isPlainObject } from './utils';
 
 export type Falsy = false | null | undefined;
 
-export function mergeChainedOptions<T, U>(params: {
+export function mergeChainedOptions<T, U = unknown, F = Falsy>(params: {
   defaults: T;
-  options?: ChainedConfig<T> | Falsy;
+  options?: ChainedConfig<T> | F;
   mergeFn?: typeof Object.assign;
+  isFalsy?: (v: T | ChainedConfig<T> | F) => v is F;
 }): T;
-export function mergeChainedOptions<T, U>(params: {
+export function mergeChainedOptions<T, U, F = Falsy>(params: {
   defaults: T;
-  options: ChainedConfigWithUtils<T, U> | Falsy;
+  options: ChainedConfigWithUtils<T, U> | F;
   utils: U;
   mergeFn?: typeof Object.assign;
+  isFalsy?: (v: T | ChainedConfigWithUtils<T, U> | F) => v is F;
 }): T;
-export function mergeChainedOptions<T, U>(params: {
+export function mergeChainedOptions<T, U, F = Falsy>(params: {
   defaults: T;
-  options: ChainedConfigCombineUtils<T, U> | Falsy;
+  options: ChainedConfigCombineUtils<T, U> | F;
   utils: U;
   mergeFn?: typeof Object.assign;
   useObjectParam?: true;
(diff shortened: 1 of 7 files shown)


9aoy:
Can you add an e2e assertion about this case (about the number of generated scripts)?

We have a related e2e case, and I think we can add an assertion as following: 
https://github.com/web-infra-dev/rsbuild/blob/0759599e6d280b72a4a7bbc293be4c15b9a5316c/e2e/cases/html/inject/index.test.ts#L16

```
expect(
  html.indexOf('/js/index') === html.lastIndexOf('/js/index'),
).toBeTruthy();
```

packages/shared/src/mergeChainedOptions.ts
```diff
@@ -7,43 +7,74 @@ import { isFunction, isPlainObject } from './utils';
...
+  isFalsy?: (v: T | ChainedConfigCombineUtils<T, U> | F) => v is F;
 }): T;
-export function mergeChainedOptions<T extends Record<string, unknown>>({
+export function mergeChainedOptions<T, U = unknown, F = Falsy>({
   defaults,
   options,
   utils,
   mergeFn = Object.assign,
   useObjectParam,
+  isFalsy = (
```
chenjiahan, line 37: Can we simplify this type declaration, complex types are difficult to maintain 😂

xc2:
> Can you add an e2e assertion about this case (about the number of generated scripts)?
> 
> We have a related e2e case, and I think we can add an assertion as following:
> 
> https://github.com/web-infra-dev/rsbuild/blob/0759599e6d280b72a4a7bbc293be4c15b9a5316c/e2e/cases/html/inject/index.test.ts#L16
> 
> ```
> expect(
>   html.indexOf('/js/index') === html.lastIndexOf('/js/index'),
> ).toBeTruthy();
> ```

added

packages/shared/src/mergeChainedOptions.ts
xc2: ya. since parameters type of implementation signature does not that matter, i've roll the complex types back to `unknown` so that we can pay more focus on the overload signatures.

chenjiahan: 👍

chenjiahan (approved):

chenjiahan:
👍👍

PR #1009, opened 2023-12-20
fix: inspect config not works in dev

9aoy:
## Summary

onBeforeStartDevServer should called before listen and after compile, so inspect config  can works.

image

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan:
any context for this change?

chenjiahan (approved):

PR #1070, opened 2023-12-26
perf(plugin-svgr): reuse SWC loader to improve perf

chenjiahan:
## Summary

Implement `svgr-loader` to replace `@svgr/webpack`:

1. Better performance, `@svgr/webpack` depends on babel for transformation, and Rsbuild can reuse the builtin swc-loader to replace babel.
2. Less dependencies, no need to depend on `@babel/preset-env` and more.

## Checklist

- [x] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/e2e/cases/svg/svgo.test.ts
+++ b/e2e/cases/svg/svgo.test.ts
@@ -51,6 +51,6 @@ test('should add id prefix after svgo minification', async () => {
   const content = readFileSync(indexJs!, 'utf-8');
 
   expect(
-    content.includes('"linearGradient",{id:"idPrefix_svg__a"}'),
+    content.includes('"linearGradient",{id:"idPrefix_svg__a"'),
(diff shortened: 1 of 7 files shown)


e2e/cases/svg/svgo.test.ts
```diff
@@ -51,6 +51,6 @@ test('should add id prefix after svgo minification', async () => {
   const content = readFileSync(indexJs!, 'utf-8');
 
   expect(
-    content.includes('"linearGradient",{id:"idPrefix_svg__a"}'),
+    content.includes('"linearGradient",{id:"idPrefix_svg__a"'),
```
chenjiahan, line 54: @svgr/webpack is compiled to React 16 JSX by default. Now Rsbuild uses SWC, so React new JSX will be used. The output are slightly different.

PR #1075, opened 2023-12-26
fix: reload when the updated modules is empty

tmkx:
## Summary

https://github.com/web-infra-dev/modern.js/pull/5152

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

PR #1076, opened 2023-12-26
Revert "fix: reload when the updated modules is empty"

tmkx:
Reverts web-infra-dev/rsbuild#1075


chenjiahan:
Note: revert because this change breaks MPA:

> When there are multiple entries, it will result in changing A to HMR, but B will reload.

chenjiahan (approved):

PR #1097, opened 2023-12-28
fix(plugin-source-build): should after babel/ts-loader/swc

9aoy:
## Summary

source build plugin depends on js/ts rule, so it should be after babel/ts-loader/swc plugin.
image

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan (approved):

chenjiahan:
It is inappropriate to rely on the uni-builder plugin names in Rsbuild.

I will design a better way to define the plugin order.

PR #1113, opened 2023-12-28
Allow the generation of either 0 or more than 1 html files for per entry

### LINKED ISSUE #1035: [Feature]: Emit multiple html files for one entry

Typically, I classify the output of my web application into two categories: local build (aka debug build) and deployment build (aka release build).

These two types differ significantly in terms of the styles, logic, file size of the output js files. However, the cost of building them together or separately is not significantly different.

In this context, we are discussing builds for different environments or markets. For instance, builds for the test environment, staging environment, production environment, China region, US region, AOC OEM, T-Mobile OEM.

Usually, people utilize the `define` options to generate different js outputs based on environment variables.

These builds have similar characteristics, but due to differences in backend URLs, product names, and copywriting, the large artifacts has only a few characters differing between them.

Hence, my usual approach involves building once and generating one assets output along with multiple HTML outputs. I include the distinct parts as `<meta>` tags or global objects with `window` in the HTML, which can then be accessed by js code at runtime.

Pros:

1. In simple terms, it involves building once and running everywhere.
2. It also improves the CDN cache hit rate for the js files, as all environments and markets share the same URL.
3. Locally, I can test my code in all environments and markets with just one build, reducing the chance of insufficient testing due to inconvenience caused by multiple builds.

let `html` or `tools.htmlPlugin` allow a function to return a set of options for the html plugin.

xc2:
## Summary

As it's hard to change `ChainedConfigWithUtils<HTMLPluginOptions>` to `ChainedConfigWithUtils<false | HTMLPluginOptions | HTMLPluginOptions[]>` without breaking developers' config, I added two new properties into`HTMLPluginOptions` instead:

- `disabled` for bypass html generation for specific entry
- `flavors` for extra html generation for specific entry

API Detail: https://deploy-preview-1113--rsbuild.netlify.app/config/tools/html-plugin#additional-properties

## Related Links

Closes #1035

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/e2e/cases/html/multiple-output/index.test.ts
+++ b/e2e/cases/html/multiple-output/index.test.ts
@@ -1,0 +1,119 @@
+import { basename, relative } from 'path';
+import { expect, test } from '@playwright/test';
+import { build } from '@scripts/shared';
+
+test.describe('tools.htmlPlugin for multiple outputs', async () => {
+  let rsbuild: Awaited<ReturnType<typeof build>> | undefined = undefined;
+  let htmls: Record<string, string> = {};
+  test.beforeAll(async () => {
+    const _rsbuild = await build({
+      cwd: __dirname,
+      runServer: true,
+      rsbuildConfig: {
+        source: {
+          entry: {
+            ibmPreOnly: './src/ibm-only.js',
+            app: './src/app.js',
+            landing: './src/landing.js',
+          },
+        },
+        tools: {
+          htmlPlugin(config, { entryName }) {
+            if (entryName === 'ibmPreOnly') {
+              return { disabled: true };
+            }
+            if (entryName === 'app') {
+              return {
+                ...config,
+                title: 'App Default',
+                filename: 'app-default.html',
+                publicPath: '/',
+                flavors: [
+                  {
+                    title: 'App - IBM',
+                    filename: 'app-ibm-oem.html',
+                    chunks: ['ibmPreOnly', ...(config.chunks as string[])],
+                    publicPath: 'https://ibm-cdn.example.com/',
+                    meta: { 'api:prefix': 'https://ibm.example.com' },
+                  },
+                  {
+                    title: 'App - AOC',
+                    filename: 'app-aoc-oem.html',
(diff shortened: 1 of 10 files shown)


chenjiahan:
Usually we should avoid extending the third-party types, as the extended properties may be conflict with the native HTMLPlugin options in the future.

Can we find some other approaches for this? 

For example, if user want to generate extra HTML files, how about just register an extra HtmlPlugin:

```js
export default defineConfig({
  tools: {
    rspack(config, { HtmlPlugin }) {
      config.plugins?.push(
        new HtmlPlugin({
          // some options
        }),
      );
    },
  },
});
```

xc2:
> Usually we should avoid extending the third-party types, as the extended properties may be conflict with the native HTMLPlugin options in the future.
> 
> Can we find some other approaches for this?
> 
> For example, if user want to generate extra HTML files, how about just register an extra HtmlPlugin:
> 
> ```js
> export default defineConfig({
>   tools: {
>     rspack(config, { HtmlPlugin }) {
>       config.plugins?.push(
>         new HtmlPlugin({
>           // some options
>         }),
>       );
>     },
>   },
> });
> ```

I'd like to have all the html files related to one entry configured in one place, instead of having one in `tools.html` and additional ones in `tools.rspack`. further more, i'd like the  extra html file options to inherit some options from the main option.

the initial proposal is:

- `() => [] | false` for 0 output
- `(options) => options` for 1 output
- `(options) => [{...options, filename: 'default.html'}, {...options, filename: 'flavor-1.html'}]` for multiple outputs

sadly the type change will break users' options as rsbuild's options are "chained" with consistent input and output.

chenjiahan:
There may be a better solution, still thinking..

xc2:
> There may be a better solution, still thinking..

nvm. the current api design of this pr is not that beautiful imo. pls feel free to share your thoughts on anything

xc2:
I'm closing this PR for a conceptualizing modern.js plugin with a browser runtime included.

chenjiahan:
👍

PR #1156, opened 2024-01-03
feat: allow passing falsy plugins

Asuka109:
## Summary

- Enable to pass falsy values as plugins.
- Warning when got webpack-like plugin object.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -19,6 +19,7 @@ export type {
   NormalizedConfig,
   // Plugin Types
   RsbuildPlugin,
+  RsbuildPlugins,
   RsbuildPluginAPI,
 } from './types';
 export type {
--- a/packages/shared/src/pluginStore.ts
+++ b/packages/shared/src/pluginStore.ts
@@ -31,7 +31,7 @@ export function createPluginStore(): PluginStore {
       }
       if (!isFunction(setup)) {
         if (isFunction(apply)) {
-          const _name = name ?? plugin.constructor?.name ?? 'SomeWebpackPlugin';
+          const _name = plugin.constructor?.name ?? 'SomeWebpackPlugin';
 
           const message =
             `${color.yellow(_name)} looks like a webpack or Rspack plugin, ` +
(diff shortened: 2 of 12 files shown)


packages/core/src/index.ts
```diff
@@ -19,6 +19,7 @@ export type {
   NormalizedConfig,
   // Plugin Types
   RsbuildPlugin,
+  RsbuildPlugins,
```
chenjiahan, line 22: why do we need to export this type in the entry module?

packages/shared/src/pluginStore.ts
```diff
@@ -3,78 +3,74 @@ import type {
   PluginStore,
   RsbuildPlugin,
   RsbuildPluginAPI,
-  BundlerPluginInstance,
+  RsbuildPlugins,
 } from './types';
 import { color, isFunction } from './utils';
 
-function validatePlugin(plugin: unknown) {
```
chenjiahan, line 10: why remove the validatePlugin function

packages/shared/src/pluginStore.ts
```diff
@@ -3,78 +3,74 @@ import type {
...
     options?: { before?: string },
   ) => {
     const { before } = options || {};
-    newPlugins.forEach((newPlugin) => {
-      validatePlugin(newPlugin);
+    for (const plugin of newPlugins) {
+      if (!plugin) continue;
+      if (typeof plugin !== 'object') {
+        throw new Error(
+          `expect plugin instance is object, but got ${typeof plugin}.`,
```
chenjiahan, line 22: the message is reverted 🌚

packages/core/src/index.ts
Asuka109: I believe that all types used by the config object should be exported. This way, upper toolkits like modern.js can extend or assemble the config type from scratch.

packages/shared/src/pluginStore.ts
Asuka109: I removed it while resolving conflicts. Should I bring it back?

packages/core/src/index.ts
chenjiahan: This file only exports **public and documented** public API, there is a comment:

image

PR #1157, opened 2024-01-03
feat: extract new function `assertPlugin` to validate format of plugins

Asuka109:
## Summary

Based on https://github.com/web-infra-dev/rsbuild/pull/1156

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/shared/src/pluginStore.ts
+++ b/packages/shared/src/pluginStore.ts
@@ -5,7 +5,72 @@ import type {
   RsbuildPluginAPI,
   RsbuildPlugins,
 } from './types';
-import { color, isFunction } from './utils';
+import { color } from './utils';
+
+class PluginTypeError extends Error {
+  constructor(plugin: unknown) {
+    super(
+      `Expect Rsbuild plugin instance to be an object, but got ${typeof plugin}.`,
+    );
+  }
+}
+
+class PluginMissingNameError extends Error {
+  constructor() {
+    super(`Rsbuild plugin should have a "name" field for identification.`);
+  }
+}
+
+class WebpackPluginLikeError extends Error {
+  constructor(plugin: object) {
+    const name = plugin.constructor?.name ?? 'SomeWebpackPlugin';
+    const message =
+      `${color.yellow(name)} looks like a webpack or Rspack plugin, ` +
+      `please use ${color.yellow('`tools.rspack`')} to register it:\n`;
+    const example = [
+      '  // rsbuild.config.ts',
+      '  export default {',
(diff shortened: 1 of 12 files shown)


packages/shared/src/pluginStore.ts
```diff
@@ -17,47 +82,9 @@ export function createPluginStore(): PluginStore {
...
-        throw new Error(
-          `Expect Rsbuild plugin.setup to be a function, but got ${typeof plugin}.`,
-        );
-      }
+      assertPlugin(plugin);
       if (plugins.some((item) => item.name === plugin.name)) {
-        logger.warn(
-          `Rsbuild plugin "${plugin.name}" registered multiple times.`,
-        );
+        logger.warn(new PluginNameDuplicatedError(plugin));
```
chenjiahan, line 87: why use logger.warn to print an error?

packages/shared/src/pluginStore.ts
```diff
@@ -5,7 +5,72 @@ import type {
...
+    const name = plugin.constructor?.name ?? 'SomeWebpackPlugin';
+    const message =
+      `${color.yellow(name)} looks like a webpack or Rspack plugin, ` +
+      `please use ${color.yellow('`tools.rspack`')} to register it:\n`;
+    const example = [
+      '  // rsbuild.config.ts',
+      '  export default {',
+      '    tools: {',
+      '      rspack: {',
+      `        plugins: [new ${name}()]`,
```
chenjiahan, line 35: No name here, you should test it.

image

chenjiahan:
You should respect the original code instead of making changes without testing. I will close this PR.

packages/shared/src/pluginStore.ts
Asuka109: It seems to be the result of unexpected merging of https://github.com/web-infra-dev/rsbuild/pull/1156. This PR is intended to fix it.

packages/shared/src/pluginStore.ts
Asuka109: I just want to extract the error initialization to make the logic of `assertPlugin` clearer. Besides being used for pluginStore, I think it can also work in other cases where type guards are needed to restrict plugin types. 
Whether or not to display a warning when showing is worth discussing. Perhaps we can consider hiding the error stack by default in logger.warn, but displaying it when `LOG_LEVEL=verbose`.
Of course, this is not something that should be done in this PR. I would prefer it to be changed here as `logger.warn('STRING')`.

PR #1166, opened 2024-01-04
feat: add new hook `onRestartDevServerHook`

Asuka109:
## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/core/src/server/devServer.ts
+++ b/packages/core/src/server/devServer.ts
@@ -231,6 +231,7 @@ export async function startDevServer<
         const closeServer = async () => {
           await devMiddlewares.close();
           httpServer.close();
+          await options.context.hooks.onRestartDevServerHook.call();
         };
 
         onBeforeRestartServer(closeServer);
--- a/.changeset/rare-suits-hear.md
+++ b/.changeset/rare-suits-hear.md
@@ -1,0 +1,6 @@
+---
+'@rsbuild/shared': minor
+'@rsbuild/core': minor
+---
+
+feat: add new hook `onRestartDevServerHook`
--- a/packages/core/src/provider/core/initHooks.ts
+++ b/packages/core/src/provider/core/initHooks.ts
@@ -15,6 +15,7 @@ import type {
   OnBeforeStartProdServerFn,
   OnAfterCreateCompilerFn,
   OnBeforeCreateCompilerFn,
+  OnRestartDevServerFn,
 } from '@rsbuild/shared';
 
 export function initHooks() {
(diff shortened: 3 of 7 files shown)


Asuka109:
!canary

packages/core/src/server/devServer.ts
```diff
@@ -231,6 +231,7 @@ export async function startDevServer<
         const closeServer = async () => {
           await devMiddlewares.close();
           httpServer.close();
+          await options.context.hooks.onRestartDevServerHook.call();
```
9aoy, line 234: this hook should not called in closeServer function, it not consistent with semantics.

Asuka109:
@9aoy @chenjiahan 
This hook is used by external plugins, such as plugin-wrangler, to clean up when restarting or shutting down the devserver.
Currently it is called at the expected timing of the lifecycle. However @9aoy noticed that the `closeServer` callback and `onRestartDevServerHook` are also called when manually closing the devserver, which does not match their intended meaning.

https://github.com/web-infra-dev/rsbuild/blob/a89f2790f54b609903413ca7ab94c3e0370f1bbf/packages/core/src/server/devServer.ts#L231-L245

By reviewing the invokers of `onBeforeRestartServer`, I found that only one other line uses this function:

https://github.com/web-infra-dev/rsbuild/blob/d353a26a17f96c1d9a2e01cc16798f786824031a/packages/core/src/loadEnv.ts#L39-L45

This seems strange because:
- The `onBeforeRestartServer` stores cleaners scoped by module and may interfere with other builder instances in the same process.
- The cleanup of `loadEnv` is currently only invoked during restart but appears to be necessary for closure as well.

So how about removing the registration of `onBeforeRestartServer` and replacing it with a new builder plugin hook named either `onCloseDevServer` or `onCleanupDevServer`?

9aoy:
> So how about removing the registration of `onBeforeRestartServer` and replacing it with a new builder plugin hook named either `onCloseDevServer` or `onCleanupDevServer`?

I think `onCloseDevServer` much better.

chenjiahan:
`loadEnv` is an independent function, is can not use rsbuild plugin hooks now.

Asuka109:
> `loadEnv` is an independent function, is can not use rsbuild plugin hooks now.

I would like to add a new callback as one of returns of `loadEnv`, and it can be able to invoked from some other places. So it will works like:

```ts
export async function init() {
  try {
    const root = process.cwd();
    const { publicVars, cleanup } = loadEnv({ cwd: root });
    const { createRsbuild } = await import('../createRsbuild');

    const builder = await createRsbuild({
      cwd: root,
      rsbuildConfig: config,
    });
    builder.onCloseDevStart(cleanup);
    return builder;
  } catch (err) {
  }
}
```

chenjiahan:
LGTM, please send separate PR for these changes

Asuka109:
> LGTM, please send separate PR for these changes

Sure, I think the `onRestartDevServerHook` is not needed anymore. I would close this PR and create a new one to work with `loadEnv` and `onCloseDevStartHook`.

PR #1170, opened 2024-01-04
chore(babel-plugin): babel-loader should not added when config is not modify

9aoy:
## Summary

the default Babel configuration will not have additional impact on the compilation results, except for increasing the build time, so we don't need add babel if user not modify / add any config.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


9aoy:
!canary

PR #1173, opened 2024-01-04
chore: turn `loadEnv` into a pure function

Asuka109:
## Summary

Turn `loadEnv` into a pure function so that we can register hook in some other places.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/core/src/cli/commands.ts
+++ b/packages/core/src/cli/commands.ts
@@ -1,10 +1,11 @@
 import { join } from 'path';
 import { existsSync } from 'fs';
-import { color, logger } from '@rsbuild/shared';
+import { color, isDev, logger } from '@rsbuild/shared';
 import { program } from '@rsbuild/shared/commander';
 import { loadEnv } from '../loadEnv';
 import { loadConfig } from './config';
 import type { RsbuildMode } from '..';
+import { onBeforeRestartServer } from 'src/server/restart';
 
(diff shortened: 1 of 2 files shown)


packages/core/src/loadEnv.ts
```diff
@@ -34,19 +33,20 @@ export function loadEnv({
...
 
-  // clear the assigned env vars before restarting dev server
-  if (isDev()) {
-    onBeforeRestartServer(() => {
-      Object.keys(parsed).forEach((key) => {
-        if (process.env[key] === parsed[key]) {
-          delete process.env[key];
-        }
-      });
+  let _cleaned = false;
```
chenjiahan, line 36: Variable name do not require an underscore prefix here:

```suggestion
  let cleaned = false;
```

packages/core/src/cli/commands.ts
```diff
@@ -42,7 +43,9 @@ export async function init({
 
   try {
     const root = process.cwd();
-    const { publicVars } = loadEnv({ cwd: root });
+    const envs = loadEnv({ cwd: root });
```
chenjiahan, line 46: `loadEnv` is an public API, please update the document: https://rsbuild.dev/api/javascript-api/core#loadenv

chenjiahan (approved):

PR #1176, opened 2024-01-04
fix(babel): presetEnv need allow null value

zllkjc:
## Summary

in [@rsbuild/babel-preset base](https://github.com/web-infra-dev/rsbuild/blob/7a5eb25fff1bfdc449f2e27487882f1ae6f66b72/packages/babel-preset/src/base.ts#L17), it seems we can pass in null value.

```typescript
if (presetEnv) {
   // some code
}
```

but actually not, if we use the web preset.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/.changeset/famous-penguins-tan.md
+++ b/.changeset/famous-penguins-tan.md
@@ -1,0 +1,6 @@
+---
+'@rsbuild/babel-preset': patch
+---
+
+fix: presetEnv need allow null value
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):

chenjiahan:
Nice catch 😄

PR #1185, opened 2024-01-05
feat(plugin-svelte): support pass preprocess options

tmkx:
## Summary

1. support pass preprocess options
2. add frequently used loader typings

## Checklist

- [x] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/plugin-svelte/src/index.ts
+++ b/packages/plugin-svelte/src/index.ts
@@ -2,6 +2,27 @@ import path from 'path';
 import { logger } from '@rsbuild/core';
 import { deepmerge } from '@rsbuild/shared';
 import type { RsbuildPlugin } from '@rsbuild/core';
+import type { CompileOptions } from 'svelte/compiler';
+import type {
+  AutoPreprocessOptions,
+  Transformer,
+} from 'svelte-preprocess/dist/types';
+
+export type { AutoPreprocessOptions, Transformer };
+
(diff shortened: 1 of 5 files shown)


packages/plugin-svelte/src/index.ts
```diff
@@ -2,14 +2,40 @@ import path from 'path';
...
    * See https://github.com/sveltejs/svelte-loader
    */
-  svelteLoaderOptions?: Record<string, any>;
+  svelteLoaderOptions?: SvelteLoaderOptions;
+  /**
+   * The options of svelte-preprocess
+   *
+   * See https://github.com/sveltejs/svelte-preprocess
+   */
+  preprocessOptions?: AutoPreprocessOptions;
```
chenjiahan, line 38: Thanks! Can you add an unit test for this new option?

see: packages/plugin-svelte/tests/index.test.ts

tmkx: done :)

but the snapshot was not serialized expectably

chenjiahan (approved):

chenjiahan:
Thanks! ❤️ 

Can you submit another PR to update the document?

packages/plugin-svelte/src/index.ts
chenjiahan: haha, maybe we can improve the snapshot in the future

PR #1220, opened 2024-01-09
types: allow to narrow Compiler type for createCompiler

chenjiahan:
## Summary

Allow to narrow Compiler type for createCompiler.

```ts
import type { Compiler } from '@rspack/core';

const compiler = rsbuild.createCompiler<Compiler>();
```

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan:
!canary

PR #1243, opened 2024-01-11
perf(plugin-svgr): prebundle url-loader to reduce dependencies

chenjiahan:
## Summary

Prebundle url-loader to reduce dependencies, this allows the plugin to reuse shared dependencies, and fix the `webpack` peer dependency warning.

## Related Links

https://github.com/webpack-contrib/file-loader/blob/master/package.json#L40-L42

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


SoonIter:
Sorry for the bother, can `type: 'asset/source',` replace the `file-loader`?
https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

chenjiahan:
No, `type: 'asset/resource'` can not be used with other loaders (such as SVGR loader)

PR #1286, opened 2024-01-15
build: clean build script, restrict hoist of pnpm

fi3ework:
## Summary

- restrict pnpm hoisted dependencies to avoid phantom deps for deps.
   - add missing dependencies after hoisting has been restricted.
   - set `cwd` for certain test cases to correctly resolve dependencies after hoisting has been restricted. (By the way, I believe all the current working directories should be set to the case directory instead of the current working directory, which is the workspace root now.)
   - remove the non-pure shims for ESM. Since the phantom dependencies are no longer available, the test case `e2e/cases/mjs-artifact/test.mjs` would fail because `global.require` will be set concurrently and fail due to the non-availability of the phantom dependency.
- fix some steps in the contribution doc.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,17 +147,17 @@ pnpm run --filter @rsbuild/some-package test
 
 In addition to the unit tests, the Rsbuild also includes end-to-end (E2E) tests, which checks the functionality of the application as a whole.
 
-You can run the `test:e2e` command to run the E2E tests:
+You can run the `e2e` command to run the E2E tests:
 
 ```sh
-pnpm run test:e2e
+pnpm run e2e
 ```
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -147,17 +147,17 @@ pnpm run --filter @rsbuild/some-package test
```
chenjiahan, line 147: This PR contains several changes, can you split them to into separate PRs? 😄

Small PR is better for reviewing and generating changelog.

fi3ework: Sure. I'm closing this PR for now.

PR #1305, opened 2024-01-16
docs: amend e2e command description

fi3ework:
## Summary

fix steps in the contribution doc

## Checklist

- [ ] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,17 +147,19 @@ pnpm run --filter @rsbuild/some-package test
 
 In addition to the unit tests, the Rsbuild also includes end-to-end (E2E) tests, which checks the functionality of the application as a whole.
 
-You can run the `test:e2e` command to run the E2E tests:
+You can run the `e2e` command to run the E2E tests:
 
 ```sh
-pnpm run test:e2e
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -147,17 +147,17 @@ pnpm run --filter @rsbuild/some-package test
...
-pnpm run test:e2e
+pnpm run e2e
 ```
 
 If you need to run a specified test, you can add keywords to filter:
 
 ```sh
-# Only run test cases with the copy-assets keyword
-npx jest copy-assets
+# Only run test cases which contains `vue` keyword with webpack
```
chenjiahan, line 159: I usually use "pnpm e2e:rspack vue", which is shorter. Also note that the Vue cases only support Rspack

fi3ework: `e2e:rspack` is a shortcut indeed. I added both path match and name match anyway.

chenjiahan (approved):

CONTRIBUTING.md
chenjiahan: 👍👍

PR #1306, opened 2024-01-16
build: remove require shims

fi3ework:
## Related Links

Part of https://github.com/web-infra-dev/rsbuild/pull/1286.

remove the non-pure shims for ESM. Since the phantom dependencies are no longer available, the test case `e2e/cases/mjs-artifact/test.mjs` would fail because `global.require` will be set concurrently and fail due to the non-availability of the phantom dependency.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan:
@10Derozan can you review this change❤️

chenjiahan (approved):

PR #1307, opened 2024-01-16
test: use test suite dirname as cwd

fi3ework:
## Summary

Part of https://github.com/web-infra-dev/rsbuild/pull/1286

I think all the cwd for test cases should be set to the case directory instead of the current working directory, which is the workspace root now.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan (approved):

chenjiahan:
LGTM 👍

Issue #1376, opened 2024-01-20
[Feature]: plugin-vue/vue2 添加 launch-editor-middleware


claneo:
### What problem does this feature solve?

用于在 vue devtool 中快速打开文件

### What does the proposed API look like?

vue-cli: 

https://github.com/vuejs/vue-cli/blob/f0f254e4bc81ed322eeb9f7de346e987e845068e/packages/%40vue/cli-service/lib/commands/serve.js#L243-L246

vite:

https://github.com/vitejs/vite/blob/639bbd65c0f7be6ad5d5d4fbe57bc15e778b7bea/packages/vite/src/node/server/index.ts#L788

chenjiahan:
Thanks, we will consider integrating this middleware in core or the Vue plugin.

PR #1381, opened 2024-01-21
feat(server): support launch editor

chenjiahan:
## Summary

Integrate https://www.npmjs.com/package/launch-editor-middleware to support launch editor.

The middleware will be enabled by default in the Rsbuild core, as many tools already use it to open the editor, see: 
 https://github.com/search?q=__open-in-editor&type=code

## Related Links

close https://github.com/web-infra-dev/rsbuild/issues/1376

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan:
@9aoy I have modified the return type of `getMiddlewares` to support passing path string.

If you have a better solution for this, feel free to adjust the implementation.

Issue #1245, opened 2024-01-11
[Feature]: Print more server logs in debug mode


chenjiahan:
### What problem does this feature solve?

When using the debug mode (DEBUG=rsbuild), we can provide more server logs for debugging.

### What does the proposed API look like?

- print all server requests.
- print the HTML fallback info.

PR #1408, opened 2024-01-23
feature(server): print more server logs in debug mode 

9aoy:
## Summary

image

img_v3_027c_8b0efd9f-ce60-4464-b0ed-5509efcd77bg

## Related Links

fix: https://github.com/web-infra-dev/rsbuild/issues/1245

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


packages/core/src/server/middlewares.ts
```diff
@@ -18,6 +18,49 @@ export const faviconFallbackMiddleware: Middleware = (req, res, next) => {
...
+    return (req, res, next) => {
+      const _startAt = process.hrtime();
+
+      const logRequest = () => {
+        const method = req.method;
+        const url: string = (req as any).originalUrl || req.url;
+        const status = Number(res.statusCode);
+
+        // get status color
+        const statusColor =
```
chenjiahan, line 34: avoid using nested ternary operator

packages/core/src/server/middlewares.ts
```diff
@@ -18,6 +18,49 @@ export const faviconFallbackMiddleware: Middleware = (req, res, next) => {
...
+export const getRequestLoggerMiddleware: () => Promise<Middleware> =
+  async () => {
+    const { default: onFinished } = await import('on-finished');
+
+    return (req, res, next) => {
+      const _startAt = process.hrtime();
+
+      const logRequest = () => {
+        const method = req.method;
+        const url: string = (req as any).originalUrl || req.url;
```
chenjiahan, line 30: no any

packages/core/package.json
```diff
@@ -56,6 +56,7 @@
     "dev": "modern build --watch"
   },
   "dependencies": {
+    "on-finished": "2.4.1",
```
chenjiahan, line 59: We do not want to add any 3rd dependency to the core package. It is recommended to implement It by ourselves. The compromise solution is prebundle

chenjiahan (approved):

PR #1427, opened 2024-01-25
feat: support mode parameter in defineConfig

Jungzl:
## Summary

support mode parameter in `defineConfig`

## Checklist

- [ ] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/packages/core/src/cli/commands.ts
+++ b/packages/core/src/cli/commands.ts
@@ -56,6 +56,7 @@ export async function init({
     const { content: config, filePath: configFilePath } = await loadConfigV2({
       cwd: root,
       path: commonOpts.config,
+      mode: commonOpts.envMode,
     });
 
     const command = process.argv[2];
--- a/packages/core/src/cli/commands.ts
+++ b/packages/core/src/cli/commands.ts
@@ -56,7 +56,7 @@ export async function init({
     const { content: config, filePath: configFilePath } = await loadConfigV2({
       cwd: root,
       path: commonOpts.config,
-      mode: commonOpts.envMode,
+      envMode: commonOpts.envMode,
     });
 
     const command = process.argv[2];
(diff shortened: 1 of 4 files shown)


packages/core/src/cli/commands.ts
```diff
@@ -56,6 +56,7 @@ export async function init({
     const { content: config, filePath: configFilePath } = await loadConfigV2({
       cwd: root,
       path: commonOpts.config,
+      mode: commonOpts.envMode,
```
chenjiahan, line 59: The mode and envMode are two different concept, so we want to distinguish between these two concepts.

- **mode**: currently uses NODE_ENV, equivalent to webpack’s mode concept, basically used to set some build configurations for a specific environment.

- **env mode**: can be set via `--env-mode <value>`, only be used to load `.env.[env-mode]` file. Can be set to any value.

chenjiahan: At the same time, I am worried that exposing envMode in rsbuild.config.ts is a potential burden for users who use the Rsbuild JavaScript API. 

They also need to pass in this value to `loadConfig` method to ensure that rsbuild.config.ts can get this value.

Jungzl: Got it, what I want to add is `envMode`

packages/core/src/cli/config.ts
```diff
@@ -115,6 +116,7 @@ export async function loadConfigByPath(configFile: string) {
       const params: ConfigParams = {
         env: getNodeEnv(),
         command,
+        envMode,
```
chenjiahan, line 119: should we use `getNodeEnv()` as the default value of `envMode`?

Adding default values can be consistent with the behavior of `loadEnv` method

chenjiahan (approved):

chenjiahan:
👍👍

PR #1453, opened 2024-01-29
fix(server): treat assetPrefix auto as ""

9aoy:
## Summary

server will slice assetPrefix when match static resource, but assetPrefix 'auto' is special.
## Related Links

https://webpack.js.org/guides/public-path/#automatic-publicpath

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/core/src/server/compilerDevMiddleware.ts
+++ b/packages/core/src/server/compilerDevMiddleware.ts
@@ -124,7 +124,7 @@ export class CompilerDevMiddleware {
         url && publicPaths.find((prefix) => url.startsWith(prefix));
 
       // slice publicPath, static asset have publicPath but html does not.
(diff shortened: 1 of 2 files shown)


packages/core/src/server/compilerDevMiddleware.ts
```diff
@@ -124,7 +124,7 @@ export class CompilerDevMiddleware {
         url && publicPaths.find((prefix) => url.startsWith(prefix));
 
       // slice publicPath, static asset have publicPath but html does not.
-      if (assetPrefix && assetPrefix !== '/') {
+      if (assetPrefix && assetPrefix !== '/' && assetPrefix !== 'auto') {
```
chenjiahan, line 127: `getPublicPathFromCompiler` already formats the publicPath 'auto' value

PR #1475, opened 2024-01-31
fix: should not depend on uni builder plugin names

chenjiahan:
## Summary

Should not depend on uni builder plugin names in Rsbuild.

This is a breaking change for Modern.js and will be released in v0.4.0.

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


chenjiahan:
!canary

PR #1480, opened 2024-01-31
chore(babel-preset): remove shared dependency

chenjiahan:
## Summary

Remove `@rsbuild/shared` dependency from `@rsbuild/babel-preset`, because babel-preset is used by many Modern.js packages, we do not want these packages to depend on `@rspack/core` (which is very large).

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/babel-preset/src/web.ts
+++ b/packages/babel-preset/src/web.ts
@@ -33,8 +33,8 @@ const getDefaultPresetEnvOption = (options: WebPresetOptions) => {
 export const getBabelConfigForWeb = (options: WebPresetOptions) => {
   if (options.presetEnv !== false) {
     options.presetEnv = {
-      ...options.presetEnv,
       ...getDefaultPresetEnvOption(options),
+      ...options.presetEnv,
     };
   }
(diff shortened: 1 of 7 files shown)


packages/babel-preset/src/web.ts
```diff
@@ -21,15 +34,15 @@ const getDefaultPresetEnvOption = (options: WebPresetOptions) => {
 };
 
 export const getBabelConfigForWeb = (options: WebPresetOptions) => {
-  const mergedOptions = deepmerge(
```
10Derozan, line 24: The action that remove deepmerge may change the behavior？

chenjiahan: The behavior is the same as before, `getDefaultPresetEnvOption` only returns these values:

image

`coreJs` can use object.assign to merge, as the properties are all required:

image

10Derozan: nice

10Derozan (approved):

PR #1495, opened 2024-02-01
fix(vue3-ts): rollback vue3-ts template env.d.ts config which will curse error on windows

southorange0929:
## Summary

vue3-ts template rollback `env.d.ts` config

Without it , it will curse error on windows.

image

## Related Links

https://github.com/vuejs/language-tools/issues/3349

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


chenjiahan:
Generally, type inference for .vue files should be provided by Volar. Is it currently not able to support Windows?

chenjiahan (approved):

chenjiahan:
Get, we'll add the type back until volar provides better support

PR #1575, opened 2024-02-18
docs: add using Rsdoctor guide

chenjiahan:
## Summary

Add using Rsdoctor guide.

## Checklist

- [ ] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/packages/document/docs/en/guide/debug/rsdoctor.mdx
+++ b/packages/document/docs/en/guide/debug/rsdoctor.mdx
@@ -4,7 +4,7 @@
 
 If you need to debug the build outputs or build process, you can use Rsdoctor for troubleshooting.
 
-## Quick start
+## Quick Start
 
 In a Rsbuild-based project, you can enable Rsdoctor as follows:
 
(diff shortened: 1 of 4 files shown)


packages/document/docs/en/guide/debug/rsdoctor.mdx
```diff
@@ -0,0 +1,49 @@
...
+# build
+RSDOCTOR=true rsbuild build
+```
+
+After running the above commands, Rsbuild will automatically register the Rsdoctor plugin, and after the build is completed, it will open the build analysis page. For complete features, please refer to [Rsdoctor document](https://rsdoctor.dev/).
+
+## Options
+
+If you need to configure the [options](https://rsdoctor.dev/config/options) provided by the Rsdoctor plugin, please manually register the Rsdoctor plugin:
+
```
9aoy, line 32: I think it would be better if the Rsdoctor plugin supports reading the `rsdoctor.config.ts` configuration file?

chenjiahan: Rsdoctor options is not very complex, it is just like other webpack plugins, so we may not need to provide a separate config file for it.

PR #1661, opened 2024-02-27
fix: progress plugin custom prefix not work

liruifengv:
## Summary

Fix progress plugin custom prefix not work.

[The docs](https://rsbuild.dev/zh/config/dev/progress-bar) says can modify the progress bar id by setting, but it's not working.

```
export default {
  dev: {
    progressBar: {
      id: 'Some Text',
    },
  },
};
```

Use `progressBar.id` as prefix if it exist.

### before
image

### after
image

## Related Links

- docs: https://rsbuild.dev/zh/config/dev/progress-bar

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/core/src/provider/plugins/progress.ts
+++ b/packages/core/src/provider/plugins/progress.ts
@@ -15,12 +15,7 @@ export const pluginProgress = (): RsbuildPlugin => ({
         return;
       }
 
-      let prefix;
-      if (options === true) {
-        prefix = TARGET_ID_MAP[target];
-      } else {
-        prefix = options.id ?? TARGET_ID_MAP[target];
-      }
+      const prefix = options !== true ? options.id : TARGET_ID_MAP[target];
 
       const { ProgressPlugin } = await import('@rspack/core');
       chain.plugin(CHAIN_ID.PLUGIN.PROGRESS).use(ProgressPlugin, [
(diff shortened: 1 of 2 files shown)


packages/core/src/provider/plugins/progress.ts
```diff
@@ -15,11 +15,12 @@ export const pluginProgress = (): RsbuildPlugin => ({
         return;
       }
 
-      const { ProgressPlugin } = await import('@rspack/core');
+      const prefix = options !== true ? options.id : TARGET_ID_MAP[target];
```
chenjiahan, line 18: Thanks for your contribution ❤️

Is `options.id` can be undefined here?

liruifengv: undefined will cause no prefix.

image

liruifengv: See rspack

image

chenjiahan: How about do it as follows, which is more intuitive:

- progressBar.id: undefined -> display the default id
- progressBar.id: '' (empty string) -> display no id

liruifengv: Great suggestion:

```ts
const prefix = (options !== true && options.id !== undefined) ? options.id : TARGET_ID_MAP[target];
```

liruifengv:
An unrelated point, your GitHub profile's Twitter link was 404. @chenjiahan

chenjiahan (approved):

chenjiahan:
LGTM 😄

chenjiahan:
> your GitHub profile's Twitter link was 404

Thanks...I have updated the link 😂

PR #1722, opened 2024-03-04
feat(plugin-source-build): support analyze exports field

9aoy:
## Summary

support analyze package.json `exports` field when use source-build plugin

## Related Links

https://rsbuild.dev/plugins/list/plugin-source-build#configuring-sub-projects

image

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


packages/monorepo-utils/src/project/project.ts
```diff
@@ -111,32 +110,18 @@ export class Project {
...
-      if (
-        typeof moduleRules === 'object' &&
-        typeof moduleRules[sourceField] === 'string'
-      ) {
-        exportsSourceDirs.push(
-          path.normalize(moduleRules[sourceField] as string),
-        );
-      }
+    if (!sourceDirs.length) {
+      throw new Error(`${this.name} 的 package.json 没有 ${sourceField} 字段`);
```
chenjiahan, line 121: English please

packages/monorepo-utils/src/project-utils/filter.ts
```diff
@@ -1,14 +1,21 @@
...
 export const filterByField =
-  (fieldName: string): FilterFunction =>
+  (fieldName: string, checkExports?: boolean): FilterFunction =>
   (projects: Project[]) => {
-    return projects.filter(p => fieldName in p.metaData);
+    return projects.filter((p) => {
+      return (
+        fieldName in p.metaData ||
+        (checkExports &&
+          getExportsSourceDirs(p.metaData.exports || {}, fieldName).length)
```
chenjiahan, line 18: Call `getExportsSourceDirs` has some performance overhead here. For example, we do not need to call `path.normalize` in this case.

chenjiahan (approved):

PR #1741, opened 2024-03-05
feat(plugin-lightningcss): integrated rsbuild with lightningcss

### LINKED ISSUE #1702: [Feature]: Lightning CSS Support

Lightning CSS is a popular CSS transformer based on Rust. It can solve the performance overhead caused by PostCSS + autoprefixer in the current Rsbuild, and it can also replace SWC CSS for CSS compression.

At present, we cannot determine whether Rspack will integrate Lightning CSS in the future, or replace SWC CSS. However, Rsbuild can first support the integration of Lightning CSS as a plugin, to verify the benefits of Lightning CSS.

> https://lightningcss.dev/

SoonIter:
## Summary

This PR is mainly implemented based on the [lightningcss-loader](https://github.com/fz6m/lightningcss-loader) and is divided into two parts 
lightningcss-loader (transform) and lightningCssMinifyWebpackPlugin (minify).

By default, `plugin-lightningcss` will replace postcss-loader with lightningcss-loader , and will replace the default rsbuild CSS minify tool (currently SwcCssMinify) with lightningCssMinifyPlugin.

If the relevant configuration is set to false, the corresponding feature will be turned off such as `transform: false` or `minify: false`

In most cases, there will be a huge performance improvement in CSS compilation.

### total
**lightningcss-loader 114ms 6x faster**
image
**postcss-loader 708ms**
image

### bootstrap-5.css
**lightningcss-loader 24ms 3.5x faster**
**postcss-loader 83ms**

### tailwind.css
**lightningcss-loader 86ms 3.5x faster**
**postcss-loader 296ms**

## Related Links

#1702 

## Checklist

- [x] plugin Tests updated.
- [x] loader e2e updated.
- [x] minify e2e updated.
- [x] Documentation updated.


packages/plugin-lightningcss/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./loader": {
+      "types": "./dist/loader.d.ts",
+      "default": "./dist/loader.js"
+    },
+    "./minimizer": {
```
chenjiahan, line 21: should `loader` and `minimizer` be a public API? maybe we can just remove these two exports fields.

packages/plugin-lightningcss/src/loader.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * modified from https://github.com/fz6m/lightningcss-loader
+ * MIT License https://github.com/fz6m/lightningcss-loader/blob/main/LICENSE
+ * Author @fz6m
+ */
+import type { LoaderContext } from 'webpack';
```
chenjiahan, line 6: can we import LoaderContext from @rspack/core

packages/plugin-lightningcss/src/types.ts
```diff
@@ -0,0 +1,38 @@
...
+
+type Implementation = typeof import('lightningcss');
+
+export type LightningCssTransformOptions = Omit<
+  TransformOptions<CustomAtRules>,
+  'filename' | 'code' | 'inputSourceMap'
+>;
+
+export type LightningCssLoaderOptions = LightningCssTransformOptions & {
+  implementation?: Implementation;
```
chenjiahan, line 11: Can we move the `implementation` field to the root of `PluginLightningcssOptions` option and add document for this field.

chenjiahan: So we do not need two `implementation` field in both `transform` and `minify`.

packages/plugin-lightningcss/src/types.ts
```diff
@@ -0,0 +1,38 @@
+import type { CustomAtRules, TransformOptions } from 'lightningcss';
+
+type Implementation = typeof import('lightningcss');
```
chenjiahan, line 3: A loose type is better here, such as `unknown`. Otherwise the user can not pass a different version of lightningcss (the type will be mismatch).

packages/document/docs/zh/plugins/list/_meta.json
```diff
@@ -16,6 +16,7 @@
   "plugin-image-compress",
   "plugin-mdx",
   "plugin-node-polyfill",
+  "plugin-lightningcss",
```
chenjiahan, line 19: the sidebar is disappeared...

https://deploy-preview-1741--rsbuild.netlify.app/plugins/list/plugin-lightningcss

image

packages/plugin-lightningcss/src/plugin.ts
```diff
@@ -0,0 +1,127 @@
...
+  browserslist: browserslistUserConfig,
+  options,
+}: {
+  chain: BundlerChain;
+  browserslist: string[];
+  utils: ModifyBundlerChainUtils;
+  options: PluginLightningcssOptions | undefined;
+}) => {
+  const defaultOptions = {
+    targets: browserslistToTargets(browserslist(browserslistUserConfig)),
```
chenjiahan, line 76: `browserslistToTargets(browserslist(browserslistUserConfig))` is called twice, we can reuse the result for both loader and minimizer.

packages/plugin-lightningcss/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export { pluginLightningcss } from './plugin';
+export { default as lightningcssLoader } from './loader';
+export { LightningCssMinifyPlugin } from './minimizer';
```
chenjiahan, line 3: when will users use the `lightningcssLoader` and `LightningCssMinifyPlugin`, can they be priviate?

packages/plugin-lightningcss/src/index.ts
```diff
@@ -0,0 +1,7 @@
...
+export type * from './types';
```
chenjiahan, line 4: Wildcard exports is not recommended to use, please only export some public types.

packages/plugin-lightningcss/tests/index.test.ts
```diff
@@ -0,0 +1,126 @@
+import { expect, describe, it } from 'vitest';
+import { pluginLightningcss, lightningcss } from '../src';
+import { createStubRsbuild } from '@scripts/test-helper';
+
+describe('plugins/lightningcss', () => {
+  it('plugin-lightningcss should replace postcss-loader with lightningcss-loader with default options', async () => {
+    const rsbuild = await createStubRsbuild({
```
chenjiahan, line 7: the `createStubRsbuild` method does not provide a default `postcss-loader`, should we use `createRsbuild` here?

packages/document/docs/zh/plugins/list/plugin-lightningcss.mdx
```diff
@@ -0,0 +1,61 @@
...
+使用 `lightningcss-loader` 替换 Rsbuild 内置的 `postcss-loader` 进行 CSS 转译，传递给 `lightningcss-loader` 的选项，请查阅 [lightningcss 仓库](https://github.com/parcel-bundler/lightningcss/blob/master/node/index.d.ts) 了解具体用法。
+
+- **类型：** `false | lightningcss.TransformOptions<CustomAtRules>`
+- **默认值：**
+
+```ts
+const defaultOptions = {
+  targets: browserslistToTargets(browserslist),
+}
+```
```
chenjiahan, line 48: we can add an example for each config option.

packages/plugin-lightningcss/src/loader.ts
```diff
@@ -0,0 +1,50 @@
...
+import { transform as _transform } from 'lightningcss';
```
chenjiahan, line 7: The `_transform` can be lazy imported, so if user use a custom implementation, we do not need to load two copies of lightningcss.

chenjiahan: but we `export { lightningcss };` in the index file, so the lightningcss will be imported by default 😂

packages/plugin-lightningcss/src/minimizer.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+import { transform as _transform } from 'lightningcss';
+import { Buffer } from 'node:buffer';
+import type { Compilation, Compiler } from 'webpack';
+import type { LightningCssMinifyPluginOptions } from './types';
+
+const PLUGIN_NAME = 'lightningcss-minify-plugin';
+const CSS_FILE_REG = /\.css(?:\?.*)?$/i;
+
+export class LightningCssMinifyPlugin {
```
chenjiahan, line 14:
```suggestion
export class LightningCSSMinifyPlugin {
```

packages/plugin-lightningcss/src/loader.ts
```diff
@@ -0,0 +1,50 @@
...
+import { Buffer } from 'node:buffer';
+import type { LightningCssLoaderOptions } from './types';
+
+const LOADER_NAME = 'lightningcss-loader';
+
+async function LightningCssLoader(
```
chenjiahan, line 13:
```suggestion
async function LightningCSSLoader(
```

packages/plugin-lightningcss/modern.config.ts
```diff
@@ -0,0 +1,3 @@
+import baseConfig from '../../scripts/modern.base.config';
```
chenjiahan, line 1: usually we need to bundle the loader file, such as https://github.com/web-infra-dev/rsbuild/blob/main/packages/plugin-pug/modern.config.ts#L6-L12

packages/plugin-lightningcss/src/plugin.ts
```diff
@@ -0,0 +1,127 @@
...
+  } satisfies LightningCssLoaderOptions;
+
+  const mergedOptions: LightningCssLoaderOptions = mergeChainedOptions({
+    defaults: defaultOptions,
+    options: options?.transform,
+  });
+
+  [CHAIN_ID.RULE.CSS, CHAIN_ID.RULE.SASS, CHAIN_ID.RULE.LESS].forEach(
```
chenjiahan, line 41: also CHAIN_ID.RULE.STYLUS

packages/plugin-lightningcss/modern.config.ts
chenjiahan: I wonder why the e2e case can pass without the `loader` bundle config.

chenjiahan: and it is recommended to use `buildConfigWithMjs` to generate mjs outputs.

packages/plugin-lightningcss/src/minimizer.ts
```diff
@@ -0,0 +1,117 @@
...
+ * MIT License https://github.com/fz6m/lightningcss-loader/blob/main/LICENSE
+ * Author @fz6m
+ */
+import { transform as _transform } from 'lightningcss';
+import { Buffer } from 'node:buffer';
+import type { Compilation, Compiler } from 'webpack';
+import type { LightningCssMinifyPluginOptions } from './types';
+
+const PLUGIN_NAME = 'lightningcss-minify-plugin';
+const CSS_FILE_REG = /\.css(?:\?.*)?$/i;
```
chenjiahan, line 12: can we use `CSS_REGEX` from `@rsbuild/shared`. / when will the `(?:\?.*)` be matched.

packages/plugin-lightningcss/src/loader.ts
chenjiahan: get 👍

chenjiahan (approved):

chenjiahan:
Awesome! We will release Rsbuild v0.5.0 in next week.

PR #1771, opened 2024-03-08
fix(core): call `onAfterBuild` on every build

xc2:
## Related Links

fixes #1765

## Checklist

- [ ] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/packages/compat/webpack/src/core/build.ts
+++ b/packages/compat/webpack/src/core/build.ts
@@ -38,9 +38,22 @@ export const build = async (
     bundlerConfigs = webpackConfigs;
   }
 
+  let isFirstCompile = true;
   await context.hooks.onBeforeBuild.call({
     bundlerConfigs: bundlerConfigs as RspackConfig[],
   });
+
+  const onDone = async (stats: Stats | MultiStats) => {
+    const p = context.hooks.onAfterBuild.call({ isFirstCompile, stats });
+    isFirstCompile = false;
+    await p;
+  };
+
+  try {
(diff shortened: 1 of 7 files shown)


chenjiahan (approved):

chenjiahan:
Nice 😄

packages/compat/webpack/src/core/build.ts
```diff
@@ -38,10 +38,23 @@ export const build = async (
...
     bundlerConfigs: bundlerConfigs as RspackConfig[],
   });
 
+  const onDone = async (stats: Stats | MultiStats) => {
+    const p = context.hooks.onAfterBuild.call({ isFirstCompile, stats });
+    isFirstCompile = false;
+    await p;
+  };
+
+  try {
```
chenjiahan, line 52: @xc2 I wonder when will `hooks.done.tapPromise` throw an error here...

chenjiahan: ok I get it, the hooks.done is a sync hook when using multiCompiler..

xc2: MultiCompiler's `done` hook is a `SyncHook`. https://github.com/webpack/webpack/blob/6842c9a2c9b70e760143188ffa5baa19a72799cb/lib/MultiCompiler.js#L63

SyncHook's tapPromise is a function that only throws

chenjiahan: yeah, this means if `onAfterBuild` is an async function, it may not work as expected when Rsbuild use MultiCompiler (has multiple targets)

Issue #1738, opened 2024-03-05
[Bug]: extra minimizer for Webpack provider


fi3ework:
### Version

```sh
N/A
```

### Details

See the reproduction (https://github.com/web-infra-dev/rspack-repro/blob/rsbuild-multiple-minimizer/rsbuild-dist/webpack.config.web.mjs#L559-L692) for a minimal Webpack provider that documented in https://github.com/web-infra-dev/rsbuild/blob/main/packages/compat/webpack/README.md#usag.

In the dumped Webpack config. Two minimizers coexist, one is Terser introduced in `compat/webpack` and SWC introduced in `compat/plugin-swc`, and they are both doing the modification process. 

I'm not familiar with modern.js and compact enough yet. Do we need to tackle this now, or is this only a minimal demo that does not represent the online projects using Webpack provider with only one minimizer? 🤔

There's been a related test case to track this.
 https://github.com/web-infra-dev/rsbuild/blob/b599cc8701c76512ac990f44862d9d7266aa0f74/packages/compat/webpack/tests/minimize.test.ts#L34-L46

### Reproduce link

https://github.com/web-infra-dev/rspack-repro/blob/rsbuild-multiple-minimizer

### Reproduce Steps

`pnpm build:rsbuild`

chenjiahan:
I think we should move the `terser-webpack-plugin` and `tools.terser` config to `@modern-js/uni-builder`, as it is only used by Modern.js.

@9aoy What do you think?

And `@rsbuild/plugin-swc` should delete the terser minimizer when applying the SWC minimizer, see https://github.com/web-infra-dev/rsbuild/blob/main/packages/compat/plugin-swc/src/plugin.ts#L115-L117 (maybe bug here)

PR #1854, opened 2024-03-18
chore(webpack): remove terser-webpack-plugin

9aoy:
## Summary

Remove `terser-webpack-plugin` and `tools.terser` configuration in webpack mode as it is only used by Modern.js (will be compatible in Modern.js uni-builder).

Once `terser-webpack-plugin` can be used in rspack mode, we will provide a js minimizer plugin to allow users to switch to `terser-webpack-plugin` for JS minification.

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/1738

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.


packages/compat/webpack/src/plugins/minimize.ts
```diff
@@ -1,27 +1,7 @@
...
-    .minimizer(CHAIN_ID.MINIMIZER.JS)
-    .use(TerserPlugin, [
-      // Due to terser-webpack-plugin has changed the type of class, which using a generic type in
-      // constructor, leading auto inference of parameters of plugin constructor is not possible, using any instead
-      finalOptions as any,
-    ])
-    .end();
-}
-
 export const pluginMinimize = (): RsbuildPlugin => ({
```
chenjiahan, line 3: this plugin can be removed

chenjiahan (approved):

PR #1899, opened 2024-03-20
feat(server): return 404 code when xxx.hot-update.json file not found

9aoy:
## Summary

Return 404 code in rsbuild dev server when `xxx.hot-update.json` file not found, so that some other frameworks which use custom server (such as modern.js) don't need to handle `.hot-update.json` file.

image

currently, in modern.js:
image

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/core/src/server/getDevMiddlewares.ts
+++ b/packages/core/src/server/getDevMiddlewares.ts
@@ -122,6 +122,16 @@ const applyDefaultMiddlewares = async ({
     upgradeEvents.push(
       compileMiddlewareAPI.onUpgrade.bind(compileMiddlewareAPI),
     );
+
+    middlewares.push((req,res, next) => {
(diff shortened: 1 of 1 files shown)


packages/core/src/server/getDevMiddlewares.ts
```diff
@@ -122,6 +122,16 @@ const applyDefaultMiddlewares = async ({
     upgradeEvents.push(
       compileMiddlewareAPI.onUpgrade.bind(compileMiddlewareAPI),
     );
+
+    middlewares.push((req,res, next) => {
```
chenjiahan, line 126:
```suggestion
    middlewares.push((req, res, next) => {
```

chenjiahan (approved):

PR #1918, opened 2024-03-22
feat(plugin-lightningcss): transform accepts boolean type

SoonIter:
## Summary

accept usages like below

```ts
pluginLightningcss({
  transform: process.env.NODE_ENV === 'development'
});
```

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/packages/document/docs/en/plugins/list/plugin-lightningcss.mdx
+++ b/packages/document/docs/en/plugins/list/plugin-lightningcss.mdx
@@ -80,6 +80,14 @@ pluginLightningcss({
 });
 ```
 
+- **Example:** Enable `lightningcss-loader` only in the development environment, still use `postcss-loader` in the production environment
+
+```ts
+pluginLightningcss({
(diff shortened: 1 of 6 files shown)


packages/document/docs/en/plugins/list/plugin-lightningcss.mdx
```diff
@@ -80,6 +80,14 @@ pluginLightningcss({
 });
 ```
 
+- **Example:** Enable `lightningcss-loader` only in the development environment, still use `postcss-loader` in the production environment
```
chenjiahan, line 83: Prefer use `production build` instead of `production environment` in the English doc:

```suggestion
- **Example:** Enable `lightningcss-loader` only in the development build, still use `postcss-loader` in the production build
```

SoonIter: done. Is there a knack to this?

chenjiahan (approved):

PR #1925, opened 2024-03-24
fix(babel-plugin): support legacy decorators out of the box

SoonIter:
## Summary

Supports legacy decorators out of the box, allowing packages such as [inversifyjs](https://github.com/inversify/InversifyJS) to compile directly.

> If you are using @babel/preset-env and legacy decorators, you must ensure the class elements transform is enabled regardless of your targets, because Babel only supports compiling legacy decorators when also compiling class properties:

https://babeljs.io/docs/babel-plugin-proposal-decorators

[babel-repl](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=AISwdgVgpgxgLgQwEYBsoAoCUAoGKEDOBABAGID25AylAE4BuIMUxA3tscaJLHOgEIJaNBkyg5OSISMbMAhAC5ig4XVlQA3NgC-2PXqA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=typescript&prettier=true&targets=&version=7.24.3&externalPlugins=%40babel%2Fplugin-proposal-decorators%407.24.1&assumptions=%7B%7D)

image

## Checklist

- [x] Tests updated.
- [ ] Documentation updated.


SoonIter:
The first commit is about the fail case

e2e/cases/decorator/2022-03/src/_decoratorBeforeExport.js
chenjiahan, line 1: can we rename the filename to `decoratorBeforeExport.js`?

packages/document/docs/en/plugins/list/plugin-babel.mdx
```diff
@@ -56,7 +56,15 @@ const defaultOptions = {
   babelrc: false,
   compact: false,
   configFile: false,
-  plugins: [],
+  plugins: [
+    ['@babel/plugin-proposal-decorators', config.source.decorators],
+    ...(isLegacyDecorators
+      ? [
+          require.resolve('@babel/plugin-transform-class-properties'),
```
chenjiahan, line 63: The default value in the document can be made easier to read.

`require.resolve` can be omitted here.

packages/plugin-babel/src/plugin.ts
```diff
@@ -66,6 +68,16 @@ export const getDefaultBabelOptions = (
         require.resolve('@babel/plugin-proposal-decorators'),
         config.source.decorators,
       ],
+      // If you are using @babel/preset-env and legacy decorators, you must ensure the class elements transform is enabled regardless of your targets, because Babel only supports compiling legacy decorators when also compiling class properties:
```
chenjiahan, line 71: If the plugin user is using `@babel/preset-env`, will this config conflict with the user config?

e2e/cases/decorator/2022-03/src/_decoratorBeforeExport.js
SoonIter: yeah, done

packages/document/docs/en/plugins/list/plugin-babel.mdx
SoonIter: done

packages/plugin-babel/src/plugin.ts
SoonIter: I think it won't
1. pluginBabel seems to be only used in rspack mode and at present it is rarely used with babel-preset.

2. For webpack, I haven't added this plugin, get error in legacy decorators
https://github.com/web-infra-dev/rsbuild/blob/b4355beee8753b621f5ad0f5434f8e85a91f8d15/packages/babel-preset/src/base.ts#L48

packages/plugin-babel/src/plugin.ts
```diff
@@ -66,6 +68,16 @@ export const getDefaultBabelOptions = (
...
+      // see https://babeljs.io/docs/babel-plugin-proposal-decorators#legacy
+      ...(isLegacyDecorators
+        ? [
+            [
+              require.resolve('@babel/plugin-transform-class-properties'),
+              { loose: true },
```
chenjiahan, line 77: should `loose` be enabled by default? 

If other plugins set `loose: false`, will it causing conflicts?

packages/plugin-babel/src/plugin.ts
chenjiahan: > at present it is rarely used with babel-preset.

Yes, it's not a common usage, but it's better if it's not broken by this change.

SoonIter: babel-preset will add this plugin automatically, I think it is smart enough
I tried use this
```js
 plugins: [
                [
                  '@babel/plugin-proposal-decorators/lib/index.js',
                  {
                    version: 'legacy'
                  }
                ],
                '@babel/plugin-transform-class-properties/lib/index.js'
              ],
              presets: [
                [
                  '@babel/preset-typescript/lib/index.js',
                  {
                    allowNamespaces: true,
                    allExtensions: true,
                    allowDeclareFields: true,
                    optimizeConstEnums: true,
                    isTSX: true
                  }
                ],
                [
                  '@babel/preset-env/lib/index.js',
                  {
                    modules: false,
                    exclude: [
                      'transform-typeof-symbol'
                    ],
                    bugfixes: true,
                    corejs: undefined,
                    targets: [
                      'Chrome >= 53'
                    ],
                    include: [
                      '@babel/plugin-transform-class-properties'
                    ]
                  }
                ]
              ],
```

packages/plugin-babel/src/plugin.ts
chenjiahan: the `loose` option should be consistent with other preset / plugins

chenjiahan: Therefore it is not a good idea to fix the value of `loose` in the code

chenjiahan (approved):

SoonIter:
thanks~

PR #1938, opened 2024-03-26
docs: update html.tags `append` introduce

9aoy:
## Summary

update `html.tags` example and  the meaning of `append` .

## Checklist

- [ ] Tests updated.
- [ ] Documentation updated.

### DIFF
--- a/packages/document/docs/en/config/html/tags.mdx
+++ b/packages/document/docs/en/config/html/tags.mdx
@@ -16,9 +16,13 @@ type HtmlTag = {
   hash?: boolean | string | ((url: string, hash: string) => string);
   /** @default true */
   publicPath?: boolean | string | ((url: string, publicPath: string) => string);
-  /** @default true */
+  /**
+   * Add this tag to the end / beginning of the original tags
+   * @default true
+   */
   append?: boolean;
   /**
+   * Whether to add tags to head
    * Enable by default only for elements that are allowed to be included in the `head` tag.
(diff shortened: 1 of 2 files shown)


packages/document/docs/zh/config/html/tags.mdx
```diff
@@ -16,10 +16,13 @@ type HtmlTag = {
   hash?: boolean | string | ((url: string, hash: string) => string);
   /** @default true */
   publicPath?: boolean | string | ((url: string, publicPath: string) => string);
-  /** @default true */
+  /**
+   * 将该标签添加到原有标签的末尾 / 开头
```
chenjiahan, line 20:
```suggestion
   * 定义当前标签相对于原有标签的插入位置。设为 `true` 时，将被插入原有标签之后；设为 `false` 时，将被插入原有标签之前。
```

packages/document/docs/en/config/html/tags.mdx
```diff
@@ -16,9 +16,13 @@ type HtmlTag = {
   hash?: boolean | string | ((url: string, hash: string) => string);
   /** @default true */
   publicPath?: boolean | string | ((url: string, publicPath: string) => string);
-  /** @default true */
+  /**
+   * Add this tag to the end / beginning of the original tags
```
chenjiahan, line 20:
```suggestion
   * Sets the insertion position of the current tag relative to the original tags.
   * If set to `true` it will be inserted after the original tags, if set to `false` it will be inserted before the original tags.
```

chenjiahan (approved):

PR #1942, opened 2024-03-26
fix(plugin-react): no chunk group of react if federation enabled

ScriptedAlchemy:
## Summary
Resolves chunk issue where excessive bloat is downloaded if federation is applied 

## Related Links

before:
image

after:
image

While more chunks are emitted, they are smaller and do not cause module duplication 

## Checklist

- [x] Tests updated.
- [ ] Documentation updated.


chenjiahan (approved):
LGTM👍 I will release a new version today

PR #1955, opened 2024-03-28
fix: onAfterBuild hook async not work

chenjiahan:
## Summary

Fix onAfterBuild hook async not work.

The MultiCompiler of webpack does not supports `done.tapPromise`, so we need to use the `done` hook of `MultiCompiler.compilers` to implement it.

## Related Links

https://github.com/web-infra-dev/rspack/pull/6058

## Checklist


chenjiahan:
!canary

chenjiahan:
!canary

PR #1958, opened 2024-03-29
fix(overlay): detect document, respect overlay option

fi3ework:
## Related Links

- detect `document` object to detemine showing overlay
- `clearOverlay` respect overlay option

## Checklist

### DIFF
--- a/packages/core/src/client/hmr/overlay.ts
+++ b/packages/core/src/client/hmr/overlay.ts
@@ -170,7 +170,7 @@ const documentAvailable = typeof document !== 'undefined';
 export function createOverlay(err: string[]) {
   if (!documentAvailable) {
     console.info(
-      'Failed to display Rsbuild overlay since document is not available, considering turning off the overlay option.',
+      'Failed to display Rsbuild overlay since document is not available, considering turning off the `dev.client.overlay` option.',
     );
     return;
   }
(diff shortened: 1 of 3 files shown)


packages/core/src/client/hmr/overlay.ts
```diff
@@ -165,12 +165,25 @@ if (customElements && !customElements.get(overlayId)) {
   customElements.define(overlayId, ErrorOverlay);
 }
 
+const documentAvailable = typeof document !== 'undefined';
+
 export function createOverlay(err: string[]) {
+  if (!documentAvailable) {
+    console.info(
+      'Failed to display Rsbuild overlay since document is not available, considering turning off the overlay option.',
```
chenjiahan, line 173:
```suggestion
      'Failed to display Rsbuild overlay since document is not available, considering turning off the `dev.client.overlay` option.',
```

chenjiahan:
!canary

packages/core/src/client/hmr/overlay.ts
fi3ework: > ```suggestion
> 
>       'Failed to display Rsbuild overlay since document is not available, considering turning off the `dev.client.overlay` option.',
> 
> ```

user may encounter this info in upper framework, the config then will be varied.

chenjiahan: Yeah, but currently all known upper frameworks retain the original `dev.client` configuration as Rsbuild. Vite will also display the of original config name, so I tend to display `dev.client.overlay` in the log.

image

chenjiahan (approved):

chenjiahan:
🎉

PR #1959, opened 2024-03-29
fix: support output childCompiler error

9aoy:
## Summary

support output childCompiler error since some error only appear in stats.childern

img_v3_029e_40b70d53-5fad-4b26-9ff5-235ca33b925g

before:
image

after:
image

## Checklist

### DIFF
--- a/packages/core/src/provider/shared.ts
+++ b/packages/core/src/provider/shared.ts
@@ -186,12 +186,22 @@ function formatErrorMessage(errors: string[]) {
 export const getAllStatsErrors = (statsData: StatsCompilation) => {
   // stats error + childCompiler error
   // only append child errors when stats error does not exist, because some errors will exist in both stats and childCompiler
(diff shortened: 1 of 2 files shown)


packages/core/src/provider/shared.ts
```diff
@@ -181,12 +183,36 @@ function formatErrorMessage(errors: string[]) {
   return `${title}\n${tip}\n${text}`;
 }
 
+export const getAllStatsErrors = (statsData: StatsCompilation) => {
+  // stats error + childCompiler error
+  // only append child errors when stats error does not exist, because some errors will exist in both stats and childCompiler
+  return statsData.errorsCount && statsData.errors?.length === 0
```
chenjiahan, line 189: Please avoid using complex ternary operator, we can use `if...else` instead.

chenjiahan (approved):

PR #1972, opened 2024-04-01
feat: support for plugin communication

chenjiahan:
## Summary

This PR introduced two new plugin API for plugin communication:

```js
const parentSymbol = Symbol('parent-api');

const pluginParent: RsbuildPlugin = {
  name: 'plugin-parent',
  setup(api) {
    api.expose<ParentAPI>(parentSymbol, {
      initial: 1,
      double: (val: number) => val * 2,
    });
  },
};

const pluginChild: RsbuildPlugin = {
  name: 'plugin-child',
  setup(api) {
    const parentAPI = api.useExposed<ParentAPI>(parentSymbol);
    expect(parentAPI?.double(parentAPI.initial)).toEqual(2);
  },
};
```

## Checklist

### DIFF
--- a/packages/core/src/provider/initPlugins.ts
+++ b/packages/core/src/provider/initPlugins.ts
@@ -63,12 +63,26 @@ export function getPluginAPI({
     );
   };
 
+  const exposed: Array<{ id: string | symbol; api: any }> = [];
+
+  const expose = (id: string | symbol, api: any) => {
+    exposed.push({ id, api });
+  };
+  const useExposed = (id: string | symbol) => {
+    const matched = exposed.find((item) => item.id === id);
+    if (matched) {
+      return matched.api;
+    }
+  };
+
   onExitProcess(() => {
     hooks.onExit.call();
   });
 
   return {
     context: publicContext,
+    expose,
+    useExposed,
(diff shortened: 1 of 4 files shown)


packages/core/src/provider/initPlugins.ts
```diff
@@ -63,12 +63,26 @@ export function getPluginAPI({
     );
   };
 
+  const exposed: Array<{ id: string | symbol; api: any }> = [];
+
+  const expose = (id: string | symbol, api: any) => {
+    exposed.push({ id, api });
```
9aoy, line 69: Should we freeze the `api` to avoid useExposed modifying it?

chenjiahan: Yes, we should do this, and the RsbuildPluginAPI object should also be readonly.

But I want to avoid introducing runtime overhead, I will add TypeScript `ReadOnly<T>` first.

chenjiahan: I reconsidered and it seems can be allowed to modify the exposed object.

For example, pluginA can expose an object that contains some properties, and maybe pluginA want pluginB to modify the properties for communication in some cases.

PR #1977, opened 2024-04-02
chore: EditLink displays Chinese in zh locale

nanianlisao:
## Summary
Keep the same as other ecological documents
## Checklist


chenjiahan (approved):
Thanks! ❤️

PR #2010, opened 2024-04-07
fix(server): unexpected dev config overwrite server config

9aoy:
## Summary

fix `server.proxy` not work when has `dev.proxy`.

## Checklist


packages/core/src/server/helper.ts
```diff
@@ -279,7 +280,17 @@ export const getDevOptions = async ({
   return {
     devServerConfig: {
```
chenjiahan, line 281: can we avoiding merging the dev and server configs?

9aoy: done

chenjiahan (approved):

PR #2016, opened 2024-04-07
feat: only inject overlay code when dev.client.overlay is enabled

9aoy:
## Summary

overlay code is only used when `dev.client.overlay` is enabled, that can reduce the bundle size in development mode.

## Checklist


packages/core/src/client/hmr/types.d.ts
```diff
@@ -3,3 +3,8 @@ declare let module: any;
 declare let __resourceQuery: string;
 
 declare let __webpack_hash__: string;
+
+interface Window {
```
chenjiahan, line 7: please avoid polluting the window object.

chenjiahan: we can export a hook from hmr.ts and let overlay.ts to use this hook

packages/core/src/client/overlay/index.ts
```diff
@@ -1,3 +1,5 @@
+import { registerOverlay } from '@rsbuild/core/client/hmr'
```
chenjiahan, line 1: Can it just be `./hmr`? So it do not reply on the resolver.

chenjiahan: 1. resolve is slower
2. may resolve to another `@rsbuild/core`

chenjiahan (approved):

PR #2021, opened 2024-04-07
feat: support apply some stats options

9aoy:
## Summary

some [stats options](https://www.rspack.dev/config/stats.html) can be worked in Rsbuild, such as stats.preset / errors / warnings / children. 

default value:

```
{
  preset: 'errors-warnings',
  children: true,
}
```

before:
image

after:
image

## Checklist

### DIFF
--- a/packages/core/src/provider/shared.ts
+++ b/packages/core/src/provider/shared.ts
@@ -223,7 +223,7 @@ export function getStatsOptions(
     } as unknown as StatsValue;
   }
 
-  return compiler.options ? compiler.options.stats as StatsValue: undefined;
+  return compiler.options ? compiler.options.stats as StatsValue : undefined;
 }
 
 export function formatStats(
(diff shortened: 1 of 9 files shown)


packages/core/src/provider/shared.ts
```diff
@@ -174,6 +175,10 @@ function formatErrorMessage(errors: string[]) {
     color.red(isTerserError ? 'Minify error: ' : 'Compile error: '),
   );
 
+  if (!errors.length) {
+    return `${title}\n${color.yellow(`For more details, please setting 'stats.errors: true' `)}`;
```
chenjiahan, line 179: Why do users need to manually set `stats.errors: true`

9aoy: not need, unless user manually  set `stats.errors: false` or `stats.preset: none`.

chenjiahan (approved):

PR #2041, opened 2024-04-09
feat(devServer): support add etag header

9aoy:
## Summary

bump webpack-dev-middleware to support etag. see: https://github.com/webpack/webpack-dev-middleware/pull/1797

fix https://github.com/web-infra-dev/rsbuild/discussions/2037

## Checklist


packages/core/src/server/compilerDevMiddleware.ts
```diff
@@ -112,6 +112,7 @@ export class CompilerDevMiddleware {
       publicPath: '/',
       stats: false,
       callbacks,
+      etag: 'weak',
```
chenjiahan, line 115: can you comment why we use `weak` here

chenjiahan (approved):

PR #2063, opened 2024-04-11
docs: fix output.inlineStyle example

Alkaidcc:
maybe 1KB here is 1024Bytes

## Checklist

### DIFF
--- a/packages/document/docs/zh/config/output/inline-styles.mdx
+++ b/packages/document/docs/zh/config/output/inline-styles.mdx
@@ -86,7 +86,7 @@ export default {
 export default {
   output: {
     inlineStyles({ size }) {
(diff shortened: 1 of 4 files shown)


packages/document/docs/zh/config/output/inline-styles.mdx
```diff
@@ -86,7 +86,7 @@ export default {
 export default {
   output: {
     inlineStyles({ size }) {
-      return size < 10 * 1000;
+      return size < 10 * 1024;
```
chenjiahan, line 89: Can we change `10KB` to `10kB` (also the English document), so we do not need to update the example

chenjiahan (approved):

chenjiahan:
Thanks ❤️

PR #2115, opened 2024-04-16
fix(plugin-assets-retry): runtime code is cut off by </script> and support es5

SoonIter:
## Summary

test(e2e): add plugin-assets-retry
fix(plugin-assets-retry): SyntaxError and runtime code is cut off by <script>

## Related Links
none

image

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #2117, opened 2024-04-16
chore(@rsbuild/shared): set logger.level in isDebug() function

SoonIter:
## Summary

https://github.com/web-infra-dev/rsbuild/blob/00a598f0095fd53cb309102757a58c89f416107b/packages/shared/src/logger.ts#L3-L8

logger.level is set too early to be affected by `process.env.DEBUG`

So it cannot be used like this

```ts
import { build } from '@rsbuild/core';

function main() {
	process.env.DEBUG = 'rsbuild';
	build();
	process.env.DEBUG = ''
}
main()
```

and in some cases of devServer e2e, we need some logs of the server

## Related Links

https://github.com/web-infra-dev/rsbuild/pull/2115

## Checklist

- [x] Tests updated not required.
- [x] Documentation updated not required

### DIFF
--- a/packages/shared/src/logger.ts
+++ b/packages/shared/src/logger.ts
@@ -11,6 +11,7 @@ export const isDebug = () => {
     return false;
   }
 
+  logger.level = 'verbose'; // support `process.env.DEBUG` in e2e
   const values = process.env.DEBUG.toLocaleLowerCase().split(',');
   return ['rsbuild', 'builder', '*'].some((key) => values.includes(key));
 };


packages/shared/src/logger.ts
```diff
@@ -11,6 +11,7 @@ export const isDebug = () => {
     return false;
```
SoonIter, line 11: We have a condition judgement here, so it will only affect the performance when `process.env.DEBUG` exists

chenjiahan (approved):

PR #2139, opened 2024-04-18
fix: failed access globalThis in legacy browsers

chenjiahan:
## Summary

Fix failed access globalThis in legacy browsers, use window instead.

## Related Links

https://caniuse.com/?search=globalThis

## Checklist


inottn:
When would `typeof window !== 'undefined'` return false? 🤔

chenjiahan:
Some special environments, such as VS Code extension or other custom JS runtimes..

inottn:
> Some special environments, such as VS Code extension or other custom JS runtimes..

get~ 😁

PR #2143, opened 2024-04-18
feat(plugin-react): enable support for profiling React in prod env

denniscual:
## Summary
Adds support for profiling React components against a production build. This is a good use case for profiling an app when running preview in localhost.

## Related Links
- [CRA Profiling](https://create-react-app.dev/docs/production-build/#profiling)

## Checklist
- [x] Documentation updated.

### DIFF
--- a/packages/plugin-react/src/react.ts
+++ b/packages/plugin-react/src/react.ts
@@ -75,5 +75,6 @@ export const applyReactProfiler = (api: RsbuildPluginAPI) => {
     // Replace react-dom with the profiling version.
     // Reference: https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977
     chain.resolve.alias.set('react-dom$', 'react-dom/profiling');
+    chain.resolve.alias.set('scheduler/tracing', 'scheduler/tracing-profiling');
   });
 };
--- a/packages/plugin-react/src/react.ts
+++ b/packages/plugin-react/src/react.ts
@@ -71,8 +71,7 @@ export const applyReactProfiler = (api: RsbuildPluginAPI) => {
     return mergeRsbuildConfig(config, enableReactProfilerConfig);
   });
 
-  api.modifyBundlerChain((chain, { isProd }) => {
-    if (!isProd) return;
+  api.modifyBundlerChain((chain) => {
     // Replace react-dom with the profiling version.
     // Reference: https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977
     chain.resolve.alias.set('react-dom$', 'react-dom/profiling');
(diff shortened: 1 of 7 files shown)


website/docs/en/plugins/list/plugin-react.mdx
```diff
@@ -145,3 +145,20 @@ pluginReact({
...
+pluginReact({
+  // Only enable the profiler when REACT_PROFILER is true, as the option will increase the build time and adds some small additional overhead.
+  enableProfiler: process.env.REACT_PROFILER === 'true',
+});
+```
+
+In your env file, define `REACT_PROFILER`. You can customize the environment variable name if desired.
+
+```env title=".env"
+REACT_PROFILER=true
```
chenjiahan, line 163: How about adding `REACT_PROFILER=true` to CLI commands instead of defining it in the env file.

```bash
REACT_PROFILER=true npx rsbuild build
```

I'm concerned that the use of .env files may lead to users applying profiles to production bundles  and release them online.

packages/plugin-react/src/react.ts
```diff
@@ -52,3 +52,28 @@ export const applyBasicReactSupport = (
...
+        },
+      },
+    };
+    return mergeRsbuildConfig(config, enableProfilerConfig);
+  });
+
+  api.modifyBundlerChain((chain) => {
+    // Replace react-dom with the profiling version.
+    // Reference: https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977
+    chain.resolve.alias.set('react-dom$', 'react-dom/profiling');
```
chenjiahan, line 77: Should we alias 'scheduler/tracing' to 'scheduler/tracing-profiling' like CRA?

https://github.com/facebook/create-react-app/blob/0a827f69ab0d2ee3871ba9b71350031d8a81b7ae/packages/react-scripts/config/webpack.config.js#L329

https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977?permalink_comment_id=4409927#gistcomment-4409927

denniscual: I tested it without scheduler alias against v17-v18 and its working without any issues. But for assurance, we can add the alias.

website/docs/en/plugins/list/plugin-react.mdx
denniscual: That would be better 👍

packages/plugin-react/src/react.ts
chenjiahan: Yes, there are still a small number of users using React 16, adding the alias can be better😄

chenjiahan (approved):

chenjiahan:
Thanks! ❤️

PR #2162, opened 2024-04-21
chore: enable biome imports sorting

fi3ework:
## Summary

Enable biome's import sort to keep consistent importing sort (see https://biomejs.dev/analyzer/import-sorting/). Most of changed files are auto fixed, `package.json` & `biome.json` are the files that needed to be reviewed indeed.
This PR changes lots of code and will pollute the blame more or less.

## Checklist


chenjiahan (approved):
I'm not sure if the import order linting will confuse developers, but it's worth a try.

chenjiahan:
When I only update some biome ignored files, the `biome check` will fail with this error:

Screenshot 2024-04-25 at 11 22 30

You can reproduce this issue by running:

```
npx biome check --apply --formatter-enabled=false --linter-enabled=false ./packages/core/compiled/sirv/index.js
```

PR #2170, opened 2024-04-22
test(e2e): use concurrent instead of serial for globContentJSON

fi3ework:
## Summary

`globContentJSON` could be written in concurrent async logic, and some test case need to be changed to prevent uncertain result in race-condition when running concurrently.

## Checklist


chenjiahan (approved):

chenjiahan:
Nice catch

PR #2086, opened 2024-04-15
feat(plugin-asset-retry): support async chunk retry

SoonIter:
## Summary

1. implement asyncChunkRetry rspack plugin based on compilation.hooks.runtimeModule https://github.com/web-infra-dev/rspack/pull/5370
2. design based on https://github.com/mattlewis92/webpack-retry-chunk-load-plugin

## Design
We will rewrite `__webpack_require__.ensureChunk` and `__webpack_require.loadScript` to achieve this feature.

1. `__webpack_require__.ensureChunk` is a way how webpack loads async chunks and it returns a promise. We intercepted the error of this Promise.
2. `__webpack_require__.loadScript` is to calculate the request url and it defines the `src` of `<script>` and we intercepted the url calculation

## Related Links

close #91

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #2200, opened 2024-04-25
fix: Trigger onFail only when the entire retry process has failed.

nanianlisao:
## Related Links

#2086 

## Checklist


SoonIter:
I think I misunderstood the meaning of onFail. 
I thought it would be triggered every time when retry failed. I'll solve this error.😢

SoonIter:
I have fixed this in #2201  and I'll add more test cases, thank you for the nice catch

nanianlisao:
> I have fixed this in #2201 and I'll add more test cases, thank you for the nice catch

fine 👍, it would be best to also pass out the original error information during a retry. I believe this is also very necessary.

chenjiahan:
Duplicated: https://github.com/web-infra-dev/rsbuild/pull/2201

Thanks ❤️

PR #2210, opened 2024-04-26
fix(assets-entry): string.endsWith compatibility issue

chenjiahan:
## Summary

`string.endsWith` is es6 API and can be replaced with `nextDomain[nextDomain.length - 1] === '/'`.

Also renamed some variables to make the bundle smaller(asyncChunkRetry.min.js 2,995 bytes -> 2,849 bytes)

## Related Links

https://caniuse.com/?search=endsWith

## Checklist


SoonIter (approved):

SoonIter:
If I could make IE disappear in the world with one click...

PR #2224, opened 2024-04-28
fix: Acquire the real address to ensure the ability to open the edito…

nanianlisao:
## Summary

By `stats.moduleIdentifier`, we can obtain the file information from the error stack, ensuring the ability to open the editor in the overlay.

Within the compile error information of rspack and webpack, file information is not included as it is with `Vite`. This absence prevents direct pinpointing and opening of the editor from the overlay.

Currently, there is hope to acquire the source file through stats.moduleIdentifier. Unfortunately, it is not possible to obtain the exact line and column, as the error stack does not contain this information, thus it is defaulted to the first line and first column."

## Checklist

### DIFF
--- a/packages/core/src/client/formatStats.ts
+++ b/packages/core/src/client/formatStats.ts
@@ -15,6 +15,19 @@ function isLikelyASyntaxError(message: string) {
   return message.includes(friendlySyntaxErrorLabel);
 }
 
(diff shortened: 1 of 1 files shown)


chenjiahan (approved):

chenjiahan:
Thanks! I will continue to improve the implementation to make it more robust

PR #2310, opened 2024-05-11
feat: support overlay display unhandled runtime errors

nanianlisao:
## Summary
support overlay display unhandled runtime errors。 

The rendering is as follows：
image

## Checklist

### DIFF
--- a/packages/core/src/client/findSourceMap.ts
+++ b/packages/core/src/client/findSourceMap.ts
@@ -1,0 +1,91 @@
+// @ts-expect-error
+import { SourceMapConsumer } from 'source-map';
+
+const fetchContent = (url: string) => fetch(url).then((r) => r.text());
+
+const findSourceMap = async (fileSource: string, filename: string) => {
+  try {
+    // Prefer to get it via filename + '.map'.
+    const mapUrl = filename + '.map';
+    return await fetchContent(mapUrl);
+  } catch (e) {
+    const mapUrl = fileSource.match(/\/\/# sourceMappingURL=(.*)$/)?.[1];
+    if (mapUrl) return await fetchContent(mapUrl);
+  }
+};
+
+// Format line numbers to ensure alignment
+const parseLineNumber = (start: number, end: number) => {
+  const digit = Math.max(start.toString().length, end.toString().length);
+  return (line: number) => line.toString().padStart(digit);
+};
+
+// Escapes html tags to prevent them from being parsed in pre tags
+const escapeHTML = (str: string) =>
+  str
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
+
+// Based on the sourceMap information, beautify the source code and mark the error lines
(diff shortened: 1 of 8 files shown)


packages/core/src/client/findSourceMap.ts
```diff
@@ -0,0 +1,86 @@
+// @ts-expect-error
+import { SourceMapConsumer } from 'source-map';
```
chenjiahan, line 2: I think import `source-map` will resolve to the Node.js version and get an error like this:

image

chenjiahan: And we do not want to add third-party dependencies to `@rsbuild/core`, unless it can be prebundled to the `compiled` folder.

nanianlisao: Yes, I tried to upgrade `source-map` to 0.7.4 and ran into some problems. It seems to be a problem with the `source-map` package itself.

Do you think it would work if I prebundled `source-map@0.5.7` into the compiled directory? If yes, I'll try to do it this way

nanianlisao: However, version `0.7.4` has a major problem that makes it unusable on the browser side. I don't think `0.5.7` should be rejected just because it's outdated, but whether it's more important to consider its usability.

I have found no good solution to this problem without changing the version.  https://github.com/mozilla/source-map/issues/459

chenjiahan: We cannot guarantee the usability of a legacy version, it may contain some bugs and unsupported features.

chenjiahan: Is source-map necessary? What will be the impact if we don’t rely on source-map

nanianlisao: It is necessary to map source files through source-map.

If there is no such function, only error stack is displayed on overlay, which is greatly compromised in terms of experience. In my opinion, the important function of pr is lost.

9aoy: Maybe we can use [source-map-js](https://www.npmjs.com/package/source-map-js).

nanianlisao: > Maybe we can use [source-map-js](https://www.npmjs.com/package/source-map-js).

@9aoy Thanks for sharing. 

I have a new problem. When I prebundler, I get an error: `__filename is not defined`.

I didn't find a good solution and had to improvise a less elegant solution.

The esm polyfill for `__dirname` is mostly Node-based. Sincerely hope to find a better solution.

chenjiahan: The prebundle is used to bundle Node.js packages, and `source-map-js` in imported in the client code, so it can not be prebundled.

chenjiahan: For the client code, I think the new code added in the PR should be imported on demand. 

In other words, when the user does not enable the runtime error overlay, these code and `source-map-js` should be bundled into the client code, otherwise it will introduce unused code and slow down the build.

nanianlisao: > In other words, when the user does not enable the runtime error overlay, these code and `source-map-js` should not be bundled into the client code, otherwise it will introduce unused code and slow down the build.

Okay, let's take a look

PR #2348, opened 2024-05-15
fix: increase priority split-by-module so it can differ from defaultVendors

JSerFeng:
fix: increase priority split-by-module so it can differ from defaultVendors

## Summary

Webpack's defaultVendors priority is -10, so the priority of defaultVendors and vendors are the same, in some case, module will be in defaultVendors chunk, not vendor chunk

## Checklist


.changeset/three-jeans-smash.md
chenjiahan, line 1: No need to add changeset in this repo 😄

chenjiahan (approved):

PR #2398, opened 2024-05-21
docs: announcing Rsbuild v0.7

chenjiahan:
## Summary

Announcing Rsbuild v0.7:

- English: https://deploy-preview-2398--rsbuild.netlify.app/community/releases/v0-7
- Chinese: https://deploy-preview-2398--rsbuild.netlify.app/zh/community/releases/v0-7

## Related Links

https://github.com/web-infra-dev/rsbuild/discussions/2271

## Checklist


website/docs/en/community/releases/v0-7.mdx
```diff
@@ -0,0 +1,196 @@
+---
+published_at: 2024-05-23 18:00:00
+---
+
+# Announcing Rsbuild v0.7
+
+> May 23, 2024
+
+![](https://assets.rspack.dev/rsbuild/rsbuild-banner-with-storybook.png)
+
```
fi3ework, line 10: This banner image style is different with previous ones that with versions. Is that on purpose? 🤔
image

chenjiahan: Yes, in this banner I want to highlight "Storybook Rsbuild" 😄 Do you think the current banner will confuse readers?

fi3ework: This break the consistency of banner style. I'd prefer to reserve the released version of the banner to conform to the article it belongs to. I think we can have two options:

1. Keep the original banner style, adding the storybook banner as a sub-banner to the storybook section.
2. Draft a new banner style, which has current version at least, and also has the highlighted features on every banner. Future banners should do the same.

chenjiahan: You are right, I have updated the doc to adopt the first option.

The second option can be even better, but requires more effort, maybe for the v1.0 release.

fi3ework (approved):

PR #2441, opened 2024-05-26
fix: allow to load empty env vars

chenjiahan:
## Summary

Allow to load empty env vars, such as empty string. This can align the behavior with CRA.

## Checklist


goamn:
This is great thanks. I had to use empty space to get past this bug 😅

PR #2474, opened 2024-05-30
fix(typed-css-modules): avoid exporting invalid `cssExports`

imranbarbhuiya:
## Summary
use declare instead of export.
Correct me if I'm wrong but I don't think cssExports is a valid export of the css module

## Checklist


chenjiahan (approved):
Thanks! ❤️

PR #2528, opened 2024-06-06
feat(server): support load SSR module

9aoy:
## Summary
update server environment api, provide some SSR related APIs for users to use:

```ts
type RsbuildDevServer = {
  ...
  environments: {
    [name: string]: {
      ...
      loadBundle: <T>(entryName: string) => Promise<T>;  // Used to load and execute stats bundle.
      getTransformedHtml: (entryName: string) => Promise<string>; //  used to get the compiled HTML template.
    }
  };
};
```

usage:

```diff
import { createRsbuild } from '@rsbuild/core';
import express from 'express';

 export const serverRender = (rsbuild, entryName) => async (req, res, next) => {
+  const { render } = await rsbuild.environments.ssr.loadBundle(entryName);
+
+  const markup = render();
+
+  const template = await rsbuild.environments.web.getTransformedHtml(entryName);
+
+  const html = template.replace(`<!--app-content-->`, markup)
+
+  res.status(200).set({ 'Content-Type': 'text/html' }).send(html)
};

export async function startDevServer() {
  const rsbuild = await createRsbuild({});

  const app = express();

  const rsbuildServer = await rsbuild.createDevServer();

+ app.get('/', serverRender(rsbuild, 'index'))

  app.use(rsbuildServer.middlewares);
  
  ....
}
```

## Related Links

https://github.com/web-infra-dev/rsbuild/discussions/2221

## Checklist


e2e/cases/server/ssr/index.test.ts
```diff
@@ -0,0 +1,45 @@
+import { rspackOnlyTest } from '@e2e/helper';
+import { expect } from '@playwright/test';
+import { startDevServer } from './scripts/server.mjs';
+
+rspackOnlyTest.only('support ssr', async ({ page }) => {
```
chenjiahan, line 5: ssr -> SSR is better

packages/core/src/server/environment.ts
```diff
@@ -0,0 +1,76 @@
...
+): Promise<T> => {
+  const { chunks, entrypoints, outputPath } = stats.toJson({
+    all: false,
+    chunks: true,
+    entrypoints: true,
+    outputPath: true,
+  });
+
+  if (!entrypoints?.[entryName]) {
+    throw new Error(`can't find ssr entry(${entryName})`);
```
chenjiahan, line 23: This API can only be used with SSR? It there any other scenarios

packages/core/src/server/environment.ts
```diff
@@ -0,0 +1,76 @@
...
+  }, []);
+
+  if (files.length === 0) {
+    throw new Error(`can't get ssr bundle by entryName(${entryName})`);
+  }
+
+  // An entrypoint should have only one entryChunk, but there may be some boundary cases
+  if (files.length > 1) {
+    throw new Error(
+      `only support load single ssr bundle entry, but got ${files.length}: ${files.join(',')}`,
```
chenjiahan, line 46: What is `bundle entry`

packages/core/src/server/devServer.ts
```diff
@@ -222,25 +223,54 @@ export async function createDevServer<
     compileMiddlewareAPI,
   });
 
+  const pwd = options.context.rootPath;
+
+  const readFileSync = (fileName: string) => {
```
chenjiahan, line 228: Can we use async fs API for performance?

packages/core/src/server/runner/legacy/asModule.js
chenjiahan, line 1: What is the motivation of `legacy` folder here

packages/shared/src/types/config/dev.ts
```diff
@@ -18,6 +18,19 @@ export type RequestHandler = (
...
   [name: string]: {
     getStats: () => Promise<Stats>;
+
+    /**
+     * Load and execute stats bundle in Server.
+     *
+     * @param entryName - relate to rsbuild source.entry
+     * @returns the return of entry module.
+     */
+    loadBundle: <T>(entryName: string) => Promise<T>;
```
chenjiahan, line 28: Does this means user need to specify the return type of `loadBundle`?

packages/core/src/server/runner/type.ts
```diff
@@ -0,0 +1,51 @@
...
+  },
+) => Record<string, any> | Promise<Record<string, any>>;
+
+export type BasicRunnerFile = {
+  path: string;
+  content: string;
+  subPath: string;
+};
+
+export enum EEsmMode {
```
chenjiahan, line 18: Typo

packages/core/src/server/environment.ts
9aoy: log problem, other scenarios are also supported.

packages/core/src/server/runner/legacy/asModule.js
9aoy: updated. This file was copied from webpack and is not a ts file, missing type.

packages/shared/src/types/config/dev.ts
9aoy: add default unknown.

packages/core/src/server/devServer.ts
9aoy: This API is used to hack the require function in chunk (supports reading files from rsbuild memfs), which needs to return synchronously.

packages/core/src/server/environment.ts
9aoy: Related rsbuild source.entry config, but as far as I know, in the bundle results, only one chunk of every entrypoint has an entry attribute of `true`.

packages/core/src/server/environment.ts
```diff
@@ -0,0 +1,76 @@
...
+  }, []);
+
+  if (files.length === 0) {
+    throw new Error(`can't get bundle by entryName(${entryName})`);
+  }
+
+  // An entrypoint should have only one entryChunk, but there may be some boundary cases
+  if (files.length > 1) {
+    throw new Error(
+      `only support load single entryChunk, but got ${files.length}: ${files.join(',')}`,
```
chenjiahan, line 46:
```suggestion
      `only support load single entry chunk, but got ${files.length}: ${files.join(',')}`,
```

chenjiahan (approved):

PR #2633, opened 2024-06-19
feat: normalize environment config

9aoy:
## Summary

normalize environment config (environment config will merge with base rsbuild config), and support get normalize environment config via `rsbuild.getNormalizedConfig({ environment })`

resolve: https://github.com/web-infra-dev/rsbuild/issues/2632

## Checklist


packages/core/src/createContext.ts
```diff
@@ -56,6 +58,22 @@ async function createContextByConfig(
   };
 }
 
+export function updateEnvironmentContext(
+  context: RsbuildContext,
+  configs: EnvironmentConfig[],
+) {
+  context.environments = configs.map((config) => ({
```
chenjiahan, line 65: `context.environments` can be an object, which is easier to use:

```js
// object
console.log(context.environments.web.entry);

// array
console.log(context.environments.find(item => item.name === 'web')?.entry);
```

packages/shared/src/types/config/index.ts
```diff
@@ -107,10 +107,15 @@ export type NormalizedConfig = DeepReadonly<{
   provider?: unknown;
   _privateMeta?: RsbuildConfigMeta;
   environments?: {
-    [name: string]: EnvironmentConfig;
+    [name: string]: DeepReadonly<EnvironmentOption>;
   };
 }>;
 
+/** The normalized Rsbuild environment config. */
+export type EnvironmentConfig = Omit<NormalizedConfig, 'environments'> & {
```
chenjiahan, line 115: `EnvironmentConfig` and `EnvironmentOption` is confusing.

I prefer using `EnvironmentConfig` and `NormalizedEnvironmentConfig`.

packages/core/src/provider/initConfigs.ts
```diff
@@ -54,6 +86,12 @@ export async function initRsbuildConfig({
   context.normalizedConfig = normalizeConfig(context.config);
   updateContextByNormalizedConfig(context, context.normalizedConfig);
 
+  const environments = getEnvironmentsConfig(context.normalizedConfig);
+
+  context.environmentConfigs = environments;
```
chenjiahan, line 91: Can we just use `context.normalizedConfig.environments.[name]`?

So we do not need to add a new `context.environmentConfigs` property.

packages/core/src/createContext.ts
9aoy: done

packages/shared/src/types/config/index.ts
9aoy: done

packages/core/src/provider/initConfigs.ts
9aoy: done

chenjiahan (approved):

PR #2639, opened 2024-06-20
fix: HMR not work on online IDEs

chenjiahan:
## Summary

Fix HMR not work on online IDEs (such as [MarsCode](https://marscode.com/)).

The default  value of`client.port` should be `window.location.port` instead of `server.port`, this allows online IDEs to proxy the ws requests to dev server correctly.

## Related Links

- MarsCode React template https://www.marscode.com/dashboard/templates

## Checklist

### DIFF
--- a/packages/core/src/server/devServer.ts
+++ b/packages/core/src/server/devServer.ts
@@ -103,7 +103,6 @@ export async function createDevServer<
   });
   const devConfig = getDevConfig({
     config,
-    port,
   });
 
   const routes = formatRoutes(
--- a/packages/core/src/server/helper.ts
+++ b/packages/core/src/server/helper.ts
@@ -252,15 +252,14 @@ export const getServerConfig = async ({
 
 export const getDevConfig = ({
   config,
-  port,
 }: {
   config: NormalizedConfig;
-  port: number;
 }): DevConfig => {
   const defaultDevConfig: DevConfig = {
     client: {
       path: HMR_SOCK_PATH,
-      port: port.toString(),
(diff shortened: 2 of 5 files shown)


packages/core/src/server/devServer.ts
```diff
@@ -103,7 +103,6 @@ export async function createDevServer<
   });
   const devConfig = getDevConfig({
     config,
-    port,
```
CPatchane, line 106: Hey @chenjiahan, this should be a breaking change for v1.
This means that the dev config will no longer use the port passed from the server config.
I have a case where the assets are served on a different port (3000) than the webpack server (3800). I configured the server with the correct port 3800, but then the HMR broke because it was using the fallback on `location.port` which returns 3000 and not 3800.
The fix is to also specify the port for `dev.client.port`, which I think is a shame, if the webpack server has a different port, the HMR should follow it, unless we say otherwise by configuring it differently.

chenjiahan: I agree that this is a breaking change, I have updated the release note and added it to breaking change list. https://github.com/web-infra-dev/rsbuild/releases/tag/v1.0.0-alpha.0

chenjiahan: I knew this change would break the websocket if client and server use different ports, and users need to set `dev.client.port`, it is expected. Vite also requires the same setup (server.hmr. clientPort).

CPatchane: Ok then, at the breaking change is now documented, thanks!

PR #2647, opened 2024-06-20
feat: support apply environment html-related config

9aoy:
## Summary

- feat: support apply environment html-related config
- feat: add htmlPaths to `context.environments`
- feat: add environment param to `api.modifyHTMLTags` context and `api.getHTMLPaths`
- fix: defaultEntry should apply to environment config instead of RsbuildConfig

```ts
const Myplugin = (): RsbuildPlugin => ({
  setup(api) {
    api.modifyBundlerChain((chain, { environment }) => {
      const { htmlPaths } = api.context.environments[environment];
      console.log(htmlPaths);
    });
    api.modifyHTMLTags({ headTags, bodyTags }, { environment }) => {
        const config = api.getNormalizedConfig({ environment });
        console.log(config);
    });
  },
});
```

## Related Links

https://github.com/web-infra-dev/rsbuild/discussions/2508

## Checklist


packages/core/src/provider/initConfigs.ts
```diff
@@ -40,28 +40,39 @@ export type InitConfigsOptions = {
...
             config as unknown as NormalizedEnvironmentConfig,
           ),
           dev,
           server,
-        },
-      ]),
+        };
+
+        if (!environmentConfig.source.entry) {
+          // @ts-expect-error
```
chenjiahan, line 60: why @ts-expect-error here

packages/core/src/provider/initConfigs.ts
```diff
@@ -40,28 +40,39 @@ export type InitConfigsOptions = {
...
+      }),
     );
   }
 
   return {
     [camelCase(rsbuildSharedConfig.output.target)]: {
       ...rsbuildSharedConfig,
+      source: {
+        ...rsbuildSharedConfig.source,
+        entry: rsbuildSharedConfig.source.entry ?? getDefaultEntry(rootPath),
```
chenjiahan, line 74: `getDefaultEntry` calls `fs`, it is better for performance to cache the result to reduce fs overhead.

packages/core/src/provider/initConfigs.ts
9aoy: Complex type issues, environmentConfig extends normalizedConfig, and it define as readonly.

chenjiahan (approved):

Issue #2620, opened 2024-06-18
[Feature]: Support multiple environment configuration


9aoy:
### What problem does this feature solve?

Enhance the current [multi-target building](https://rsbuild.dev/config/output/targets) capabilities, users can define completely different Rsbuild configurations for different environments.

Different environments can be different usage scenarios of the **same rsbuild target** (node ​​SSR / node RSC) or **different rsbuild targets** (web /  node SSR)

before:
<img width="786" alt="image" src="https://github.com/web-infra-dev/rsbuild/assets/22373761/c1e783dc-5727-46d1-b04a-c5ff3e76ae29">

after:
<img width="1144" alt="image" src="https://github.com/web-infra-dev/rsbuild/assets/22373761/bfb1c446-5bd0-4687-9288-5a5c563e58b5">

### What does the proposed API look like?

> The design of environment configuration in Rsbuild is inspired by Vite Environment API. Thanks to the Vite team ❤️

Add the `environments` field in Rsbuild config, and environment configuration can override the basic Rsbuild configuration.

```title=rsbuild.config.ts
export default {
  // Shared configuration for all environments
  source: {
    alias: {
      '@common': './src/common',
    },
  },
  environments: {
    // configuration for web
    web: {
      source: {
        entry: {
          index: './src/index.client.js',
        }
      },
      target: 'web',
      tools: {
        bundlerChain: (chain) => {
           chain.plugin('xxx').use(xxx);
        }
      },
    },
    // configuration for ssr
    'ssr': {
      source: {
        entry: {
          index: './src/index.server.js',
        }
      },
       target: 'node',
       tools: {
          htmlPlugin: false,
       }
    },
  },
};
```
```[tasklist]
### MVP
- [ ] https://github.com/web-infra-dev/rsbuild/issues/2623
- [ ] https://github.com/web-infra-dev/rsbuild/issues/2632
- [x] chore: Generate rspack config based on environment instead of `output.targets`
- [x] breaking change:  Replace `output.targets` with `output.target`
- [x] breaking change:  Replace api.context.targets / entry / distPath / xxx with api.environments[index].xxx
- [x] feat: Should use environment.config instead of `api.getNormalizedConfig()` in plugins
- [x] feat: `api.getHTMLPaths` support get specify environment htmlPath by environment parameter
- [x] breaking change: `output.emitAssets` changed to boolean type
- [x] breaking change: `output.overrideBrowserslist` no longer supports object type
```

```[tasklist]
[...]

PR #2680, opened 2024-06-24
refactor!: update api.context, remove `entry`, `targets` properties

9aoy:
## Summary

Remove `api.context.entry` and `api.context.targets`,  use `api.environment` instead.

 The `distPath` and `tsconfigPath` may be different in different environments. Rsbuild will try to get the parent distPath of environments as `context.distPath`, it is recommended to use `api.environment[name]` instead of `api.context`.

```diff
const pluginFoo = () => ({
  setup(api) {
      api.modifyBundlerChain((chain, { environment }) => {
-            const { entry,  targets, tsconfigPath, distPath } = api.context;
+            const { entry,  target, tsconfigPath, distPath } = api.environments[environment];
    });
  },
});

```

## Related Links

## Checklist


packages/core/src/plugins/cache.ts
```diff
@@ -126,8 +132,8 @@ export const pluginCache = (): RsbuildPlugin => ({
         // The default cache name of webpack is '${name}-${env}', and the `name` is `default` by default.
         // We set cache name to avoid cache conflicts of different targets.
         name: useDigest
-          ? `${target}-${env}-${getDigestHash(cacheConfig.cacheDigest!)}`
-          : `${target}-${env}`,
+          ? `${target}-${env}-${environment}-${getDigestHash(cacheConfig.cacheDigest!)}`
```
chenjiahan, line 135: Maybe `target` can be removed

packages/core/src/createContext.ts
```diff
@@ -165,11 +162,14 @@ export function updateContextByNormalizedConfig(
...
 ) {
-  context.distPath = getAbsoluteDistPath(context.rootPath, config);
-
-  if (config.source.entry) {
-    context.entry = getEntryObject(config, 'web');
-  }
+  // Try to get the root distPath of context.environments as context.distPath
+  context.distPath = Object.values(context.environments).reduce((p, e) => {
+    // foo/dist/server vs foo/dist
+    if (!p || p.startsWith(e.distPath)) {
```
chenjiahan, line 168: Can this function get the common parent directory as expected?

For example, `/path/to/dist/client` and `/path/to/dist/server`, the parent directory is `/path/to/dist`

9aoy: but if `/path/dist-client` and `/path/dist-server`, the parent directory is `/path`, is expected?

chenjiahan: Yes, because we can't use `/path/dist-client' or `/path/dist-server'.

chenjiahan (approved):

colinaaa:
I cannot find `api.environments`(or `api.context.environments`) in Rsbuild `1.0.0-alpha`.

Seems like it has been changed to `utils.environments` in https://github.com/web-infra-dev/rsbuild/pull/2721.

But the `environements` does not exist in the utils of `api.onDevCompileDone`.

So, how could I migrate the following code:

```js
api.onDevCompileDone(async ({ isFirstCompile, stats }) => {
  const entries = Object.keys(api.context.entry)
})
```

chenjiahan:
Yeah we have changed the design and no longer provides "api.context.environments".

The onDevCompileDone callback should have a environements param, @9aoy do we missing it

9aoy:
> So, how could I migrate the following code:
> 
> ```js
> api.onDevCompileDone(async ({ isFirstCompile, stats }) => {
>   const entries = Object.keys(api.context.entry)
> })
> ```

@colinaaa  The environment api is not fully complete yet (also doc and features). And it seems that the `environments` parameter is missing in `onDevCompileDone` hook.
When completed, it should be migrated as follows:

```
api.onDevCompileDone(async ({ isFirstCompile, stats, environments }) => {
  const entries = Object.keys(environments).map(e => e.entry)
})
```

PR #2693, opened 2024-06-25
fix: rspack profile outputPath unexpected

9aoy:
## Summary

should not get `api.context.distPath` before config init

## Related Links

## Checklist

### DIFF
--- a/packages/core/src/plugins/rspackProfile.ts
+++ b/packages/core/src/plugins/rspackProfile.ts
@@ -43,10 +43,7 @@ export const pluginRspackProfile = (): RsbuildPlugin => ({
     }
 
     const timestamp = Date.now();
-    const profileDir = path.join(
-      api.context.distPath,
-      `rspack-profile-${timestamp}`,
-    );
+    const profileDirName = `rspack-profile-${timestamp}`;
 
     let profileSession: inspector.Session | undefined;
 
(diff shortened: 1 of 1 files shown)


packages/core/src/plugins/rspackProfile.ts
```diff
@@ -88,6 +86,10 @@ export const pluginRspackProfile = (): RsbuildPlugin => ({
     api.onBeforeStartDevServer(onStart);
 
     api.onAfterBuild(async ({ stats }) => {
+      const profileDir = path.join(api.context.distPath, profileDirName);
+
+      const loggingFilePath = path.join(profileDir, 'logging.json');
```
chenjiahan, line 91: The join calls can be merged:

```suggestion
      const loggingFilePath = path.join(api.context.distPath, profileDirName , 'logging.json');
```

packages/core/src/plugins/rspackProfile.ts
```diff
@@ -102,6 +106,9 @@ export const pluginRspackProfile = (): RsbuildPlugin => ({
       if (enableProfileTrace) {
         rspack.experimental_cleanupGlobalTrace();
       }
+      const profileDir = path.join(api.context.distPath, profileDirName);
```
chenjiahan, line 109: same

packages/core/src/plugins/rspackProfile.ts
```diff
@@ -59,11 +56,12 @@ export const pluginRspackProfile = (): RsbuildPlugin => ({
     const enableLogging =
       RSPACK_PROFILE === 'ALL' || RSPACK_PROFILE.includes('LOGGING');
 
-    const traceFilePath = path.join(profileDir, 'trace.json');
-    const cpuProfilePath = path.join(profileDir, 'jscpuprofile.json');
-    const loggingFilePath = path.join(profileDir, 'logging.json');
-
     const onStart = () => {
+      // can't get precise api.context.distPath before config init
+      const profileDir = path.join(api.context.distPath, profileDirName);
```
chenjiahan, line 61: same

packages/core/src/plugins/rspackProfile.ts
9aoy: `profileDir` is also used below.

chenjiahan (approved):

PR #2707, opened 2024-06-26
feat: add modifyEnvironmentConfig hook

9aoy:
## Summary

support modify environment config via `modifyEnvironmentConfig` hook.

Type:
```ts
type ModifyEnvironmentConfigUtils = {
  /** environment name. */
  name: string;
  /** Merge multiple Rsbuild environment config objects into one. */
  mergeEnvironmentConfig: (
    config: MergedEnvironmentConfig,
    ...configs: EnvironmentConfig[]
  ) => MergedEnvironmentConfig;
};

function ModifyEnvironmentConfig(
  callback: (
    config: MergedEnvironmentConfig,
    utils: ModifyEnvironmentConfigUtils,
  ) => MaybePromise<MergedEnvironmentConfig | void>,
): void;
```

Example: 

```ts
const myPlugin = () => ({
  setup: (api) => {
    api.modifyEnvironmentConfig((config, { name }) => {
      if (name !== 'web') {
          return;
       }
      config.html ||= {};
      config.html.title = 'My Default Title';
    });
  },
});
```
## Related Links

## Checklist


packages/core/src/provider/initConfigs.ts
```diff
@@ -107,11 +132,35 @@ export async function initRsbuildConfig({
...
     normalizeBaseConfig,
     context.rootPath,
   );
 
+  const {
+    dev: { assetPrefix, lazyCompilation, ...rsbuildSharedDev },
+    server,
+  } = normalizeBaseConfig;
+
+  for await (const [name, config] of Object.entries(mergedEnvironments)) {
```
chenjiahan, line 147: unused await?

packages/core/src/provider/initConfigs.ts
```diff
@@ -33,16 +34,33 @@ async function modifyRsbuildConfig(context: InternalContext) {
...
 
+async function modifyEnvironmentConfig(
+  context: InternalContext,
+  config: MergedEnvironmentConfig,
+  name: string,
+) {
+  logger.debug(`modify Rsbuild environment(${name}) config`);
+  const [modified] = await context.hooks.modifyEnvironmentConfig.call(config, {
+    name,
+    // @ts-expect-error
```
chenjiahan, line 45: Add the reason for @ts-expect-error
```suggestion
    // @ts-expect-error some reason
```

packages/shared/src/types/config/index.ts
```diff
@@ -97,6 +97,18 @@ export interface RsbuildConfig extends EnvironmentConfig {
   _privateMeta?: RsbuildConfigMeta;
 }
 
+export type MergedEnvironmentConfig = {
```
chenjiahan, line 100: I think the only different of `MergedEnvironmentConfig` and `NormalizedEnvironmentConfig` is `DeepReadonly<>`?

Can we just merge the two types? We do not want the config types to become so complex.

packages/shared/src/types/hooks.ts
```diff
@@ -86,11 +91,26 @@ export type ModifyRsbuildConfigUtils = {
   mergeRsbuildConfig: (...configs: RsbuildConfig[]) => RsbuildConfig;
 };
 
+export type ModifyEnvironmentConfigUtils = {
+  /** environment name. */
+  name: string;
+  /** Merge multiple Rsbuild environment config objects into one. */
+  mergeEnvironmentConfig: (
```
chenjiahan, line 98: Can we just name it `mergeRsbuildConfig`?

The implementation is `mergeEnvironmentConfig`, and `environments` is just part of Rsbuild config.

packages/shared/src/types/config/index.ts
9aoy: `MergedEnvironmentConfig` only has configs that the user can modify, but `NormalizedEnvironmentConfig` has all configurations that are read-only.


`NormalizedEnvironmentConfig` = DeepReadonly<`MergedEnvironmentConfig`  + `SharedRsbuildConfig` (server / _privateMeta / ... )>

packages/shared/src/types/hooks.ts
9aoy: But user can't modify the shared rsbuild config(not works) via `mergeRsbuildConfig`, and merge environment config  via `mergeRsbuildConfig ` is not make sense.  

It likes below:

```ts
          api.modifyEnvironmentConfig((config, { name, mergeRsbuildConfig }) => {
            return mergeRsbuildConfig(config, {
              environments: {
               [name]: {
                  source: {
                    alias: {
                      '@common1': './src/common1',
                    },
                  },
                },
              },
            });
          });
```

chenjiahan: get

chenjiahan (approved):

PR #2713, opened 2024-06-27
fix: print and serve multiple environments html route correctly

9aoy:
## Summary

print and serve multiple environments html route correctly in basic case.

When we need to serve multiple environment products, rsbuild server can run correctly in the following case: 
```ts
export default defineConfig({
  plugins: [pluginReact()],
  environments: {
    web: {},
    web1: {
      dev: {
        // assetPrefix is necessary when generating multiple environment web products, file search conflicts will occur if assetPrefix is ​​not added.
        assetPrefix: 'auto',
      },
      source: {
        entry: {
          main: './src/web1.tsx',
        },
      },
      output: {
        distPath: {
          root: 'dist/web1',
        },
      },
    },
  },
});

```

At this point, the print url is as follows:
```bash
  - index    http://localhost:3000/
  - main     http://localhost:3000/web1/main
```

Warning: not support serve multiply environments when distPath is different (non-inclusive relationship), such as, `distPath.root: 'dist' ` in environment web and `distPath.root: 'dist1' `  in environment web1 

## Related Links

## Checklist

### DIFF
--- a/packages/core/src/server/helper.ts
+++ b/packages/core/src/server/helper.ts
@@ -2,6 +2,7 @@ import type { IncomingMessage } from 'node:http';
 import net from 'node:net';
 import type { Socket } from 'node:net';
 import os from 'node:os';
+import { join, relative } from 'node:path';
 import { color } from '@rsbuild/shared';
 import type {
   NormalizedConfig,
@@ -13,6 +14,7 @@ import type {
 import { DEFAULT_DEV_HOST, DEFAULT_PORT } from '../constants';
 import { isFunction } from '../helpers';
 import { logger } from '../logger';
+import type { InternalContext } from '../types';
 
 /**
  * It used to subscribe http upgrade event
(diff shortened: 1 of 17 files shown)


packages/core/src/server/helper.ts
```diff
@@ -47,6 +49,29 @@ const formatPrefix = (prefix: string | undefined) => {
   return `${hasLeadingSlash ? '' : '/'}${prefix}${hasTailSlash ? '' : '/'}`;
 };
 
+export const getRoutes = (options: { context: InternalContext }): Routes => {
+  return Object.entries(options.context.environments).reduce<Routes>(
+    (prev, [name, context]) => {
```
chenjiahan, line 54: Rename to `environmentContext` to be more clear and avoid conflict.

```suggestion
    (prev, [name, environmentContext]) => {
```

packages/core/src/server/helper.ts
```diff
@@ -47,6 +49,29 @@ const formatPrefix = (prefix: string | undefined) => {
   return `${hasLeadingSlash ? '' : '/'}${prefix}${hasTailSlash ? '' : '/'}`;
 };
 
+export const getRoutes = (options: { context: InternalContext }): Routes => {
```
chenjiahan, line 52: destruct `context` from options to make the code shorter.

```suggestion
export const getRoutes = ({ context } : { context: InternalContext }): Routes => {
```

packages/core/src/server/helper.ts
```diff
@@ -47,6 +49,29 @@ const formatPrefix = (prefix: string | undefined) => {
...
+  return Object.entries(options.context.environments).reduce<Routes>(
+    (prev, [name, context]) => {
+      const distPrefix = relative(
+        options.context.distPath,
+        options.context.environments[name].distPath,
```
chenjiahan, line 57: Equal to environmentContext

```suggestion
        environmentContext.distPath,
```

chenjiahan (approved):

PR #2714, opened 2024-06-27
feat: add basic server environment api

9aoy:
## Summary

add server environment api.

```ts
type RsbuildDevServer = {
  ...
  environments: {
    [name: string]: {
      getStats: () => Promise<Stats>;
      loadBundle: <T>(entryName: string) => Promise<T>; // TODO
      getTransformedHtml: (entryName: string) => Promise<string | undefined>; // TODO
    }
  };
};
```
## Related Links

#2713 

## Checklist

### DIFF
--- a/packages/core/src/server/devServer.ts
+++ b/packages/core/src/server/devServer.ts
@@ -132,14 +132,14 @@ export async function createDevServer<
 
   let outputFileSystem: Rspack.OutputFileSystem = fs;
 
-  let _stats: Stats[];
+  let lastStats: Stats[];
 
   // should register onDevCompileDone hook before startCompile
   const waitFirstCompileDone = runCompile
     ? new Promise<void>((resolve) => {
         options.context.hooks.onDevCompileDone.tap(
           ({ stats, isFirstCompile }) => {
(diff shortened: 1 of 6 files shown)


packages/core/src/server/devServer.ts
```diff
@@ -197,11 +220,26 @@ export async function createDevServer<
...
   });
 
+  const environmentAPI = Object.fromEntries(
+    Object.keys(options.context.environments).map((name, index) => {
+      return [
+        name,
+        {
+          getStats: async () => {
+            await waitFirstCompileDone;
+            return _stats[index];
```
chenjiahan, line 230: Should we throw an error when `runCompile` is false? The return type is not stats

packages/core/src/server/devServer.ts
```diff
@@ -127,6 +132,24 @@ export async function createDevServer<
 
   let outputFileSystem: Rspack.OutputFileSystem = fs;
 
+  let _stats: Stats[];
```
chenjiahan, line 135: How about `lastStats` or `currentStats`.

```suggestion
  let lastStats: Stats[];
```

chenjiahan (approved):

PR #2738, opened 2024-07-01
feat: support output environment config when inspect config

9aoy:
## Summary

before:
image

after:
single environment configuration:
image

multiple environment configuration:
image

## Related Links

#2620

## Checklist


packages/core/src/config.ts
```diff
@@ -416,7 +496,17 @@ export async function outputInspectConfigFiles({
       label: 'Rsbuild Config',
       content: rawRsbuildConfig,
     },
-    ...bundlerConfigs.map(({ name, content }) => {
+    ...rawEnvironmentConfigs.map(({ name, content }) => {
+      const outputFile = `rsbuild.environment-${name}.config.mjs`;
```
chenjiahan, line 500:
```suggestion
      const outputFile = `rsbuild.${name}.config.mjs`;
```

packages/core/src/provider/provider.ts
```diff
@@ -68,7 +68,7 @@ export const rspackProvider: RsbuildProvider = async ({
     },
 
     async inspectConfig(inspectOptions) {
-      const { inspectConfig } = await import('./inspectConfig');
+      const { inspectConfig } = await import('./inspectConfig');
```
chenjiahan, line 71:
```suggestion
      const { inspectConfig } = await import('./inspectConfig');
```

chenjiahan: if your git hooks not work

packages/core/src/config.ts
```diff
@@ -391,16 +394,93 @@ export async function loadConfig({
...
+    ...rsbuildConfig,
+    pluginNames,
+  };
+
+  const rawRsbuildConfig = stringifyConfig(
+    rsbuildDebugConfig,
+    inspectOptions.verbose,
+  );
+
+  const { environmentDebugConfigs, rawEnvironmentConfigs } = Object.entries(
```
chenjiahan, line 430: I think we do not need to use `Object.entries().reduce()` loop here.

Just declare the variables and assign to it, the code will be much simpler.

- before:

```ts
const { environmentDebugConfigs, rawEnvironmentConfigs } = Object.entries(
  environments,
).reduce<{
  environmentDebugConfigs: Record<
    string,
    NormalizedEnvironmentConfig & {
      pluginNames: string[];
    }
  >;
  rawEnvironmentConfigs: Array<{
    name: string;
    content: string;
  }>;
}>(
  (prev, [name, config]) => {
    const debugConfig = {
      ...config,
      pluginNames,
    };
    prev.rawEnvironmentConfigs.push({
      name,
      content: stringifyConfig(debugConfig, inspectOptions.verbose),
    });
    prev.environmentDebugConfigs[name] = debugConfig;
    return prev;
  },
  {
    environmentDebugConfigs: {},
    rawEnvironmentConfigs: [],
  },
);
```

- after:

```ts
const environmentConfigs: Record<
  string,
  NormalizedEnvironmentConfig & { pluginNames: string[] }
> = {};

const rawEnvironmentConfigs: Array<{ name: string; content: string }> = [];

Object.entries(environments).forEach(([name, config]) => {
  const debugConfig = {
    ...config,
    pluginNames,
  };
  prev.rawEnvironmentConfigs.push({
    name,
    content: stringifyConfig(debugConfig, inspectOptions.verbose),
  });
  prev.environmentConfigs[name] = debugConfig;
});
```

packages/shared/src/types/provider.ts
```diff
@@ -42,10 +42,17 @@ export type InspectConfigOptions = {
 export type InspectConfigResult<B extends 'rspack' | 'webpack' = 'rspack'> = {
   rsbuildConfig: string;
   bundlerConfigs: string[];
+  environmentConfigs: string[];
   origin: {
-    rsbuildConfig: NormalizedConfig & {
+    rsbuildConfig: Omit<NormalizedConfig, 'environments'> & {
       pluginNames: string[];
     };
+    environmentConfigs: Record<
```
chenjiahan, line 50: Can you update the document of this API.

chenjiahan (approved):

Issue #2747, opened 2024-07-01
[Bug]: When defining dev.assetPrefix to a full URL (http://localhost:3010/subpath/) static files can't be found under said subpath


nate-summercook:
### Version

```sh
System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 7.46 GB / 32.00 GB
    Shell: 6.21.00 - /bin/tcsh
  Browsers:
    Brave Browser: 125.1.66.115
    Chrome: 126.0.6478.127
    Edge: 126.0.2592.81
    Safari: 17.5
```

### Details

I have a rather complicated setup with multiple module federation apps running in parallel. We have start script where you select which apps to start locally and which to proxy in from a deployment (very handy when working with MF). So much for the introduction.
Now because of said setup, we need to define the `dev.assetPrefix` path to a full URL (like i.e. http://localhost:3010/subpath/). Because if we only set an absolute URL, when trying to fetch async packages from one of the MF modules, it wouldn't know in which app to search (it would search it in the host app, where it's most probably not).
I noticed that when I set an absolute URL, it will make sure that the static files can be accessed through this absolute URL. Not so much if the assetPrefix is a full URL. The script tag points to the right place, but the JS files will still be located outside the subpath.

If there's another way to achieve this, then I'm happy, but currently, I'm really struggling to get this usecase (different subpath isntead of every app on root) to work.

### Reproduce link

https://github.com/nate-summercook/rsbuild-asset-prefix

### Reproduce Steps

I guess this could be reproduced without a module federation setup, it's probably a general limitation of rsbuild, that static files won't be accessible on a subpath if the `dev.assetPath` is a full URL. It's probably also a desired behavior, but it would be nice to have a way to deal with this. Searched the documentation up and down but didn't find anything.

1. Checkout my repro repository
2. run `npm i`
3. run `npm start`
4. If you change assetPrefix to /subpath/, it will work, but that doesn't help me in my module federation situation.

chenjiahan:
@9aoy can you help to check this?

9aoy:
I think this is a scenario that rsbuild misses, let me figure out how to support this.

PR #2754, opened 2024-07-02
fix(server): should match resource correctly with full url assetPrefix

9aoy:
## Summary

should match resource correctly with full url assetPrefix, like below:
```ts
export default defineConfig({
  dev: {
    assetPrefix: 'http://localhost:3000/subpath/,
  },
});
```

## Related Links

fix: #2747

## Checklist

### DIFF
--- a/packages/core/src/server/compilerDevMiddleware.ts
+++ b/packages/core/src/server/compilerDevMiddleware.ts
@@ -6,6 +6,7 @@ import type {
   DevMiddlewareAPI,
 } from './devMiddleware';
 import { SocketServer } from './socketServer';
+import { pathParse } from './helper';
 
 type Options = {
   publicPaths: string[];
@@ -122,6 +123,8 @@ export class CompilerDevMiddleware {
       etag: 'weak',
     });
 
+    const assetPrefixes = publicPaths.map(pathParse);
+
     const warp = async (
       req: IncomingMessage,
       res: ServerResponse,
@@ -129,7 +132,7 @@ export class CompilerDevMiddleware {
     ) => {
       const { url } = req;
       const assetPrefix =
-        url && publicPaths.find((prefix) => url.startsWith(prefix));
+        url && assetPrefixes.find((prefix) => url.startsWith(prefix));
 
       // slice publicPath, static asset have publicPath but html does not.
       if (assetPrefix && assetPrefix !== '/') {
--- a/packages/core/src/server/compilerDevMiddleware.ts
+++ b/packages/core/src/server/compilerDevMiddleware.ts
@@ -1,11 +1,11 @@
 import type { IncomingMessage, ServerResponse } from 'node:http';
 import type { Socket } from 'node:net';
 import type { DevConfig, NextFunction, ServerConfig } from '@rsbuild/shared';
+import { pathnameParse } from '../helpers/path';
 import type {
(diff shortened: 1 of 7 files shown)


packages/core/src/server/prodServer.ts
```diff
@@ -165,8 +165,10 @@ export async function startProdServer(
     {
       pwd: context.rootPath,
       output: {
-        path: config.output.distPath.root || ROOT_DIST_DIR,
-        assetPrefix: config.output.assetPrefix,
+        path: context.distPath,
+        assetPrefixes: Object.values(
+          context.normalizedConfig?.environments || [],
```
chenjiahan, line 170: Can we use `context.environments` here

chenjiahan: `[]` should be `{}`

packages/core/src/server/helper.ts
```diff
@@ -50,6 +50,14 @@ const formatPrefix = (prefix: string | undefined) => {
   return `${hasLeadingSlash ? '' : '/'}${prefix}${hasTailSlash ? '' : '/'}`;
 };
 
+export const pathParse = (publicPath: string): string => {
```
chenjiahan, line 53: Can be moved to `packages/core/src/helpers/path.ts`

We will move more helpers from `helper.ts` to `helpers/[name].ts` to avoid maintaining a large file.

packages/core/src/server/compilerDevMiddleware.ts
```diff
@@ -122,14 +123,16 @@ export class CompilerDevMiddleware {
...
+
     const warp = async (
       req: IncomingMessage,
       res: ServerResponse,
       next: NextFunction,
     ) => {
       const { url } = req;
       const assetPrefix =
-        url && publicPaths.find((prefix) => url.startsWith(prefix));
+        url && assetPrefixes.find((prefix) => url.startsWith(prefix));
```
chenjiahan, line 135: Maybe several assetPrefixes can match the same url?

- assetPrefixA: `/foo/bar`
- assetPrefixB: `/foo`

packages/core/src/server/prodServer.ts
9aoy: `output.assetPrefix` is not in `context.environments`, and it's not commonly used.

packages/core/src/server/compilerDevMiddleware.ts
9aoy: It's possible. Users are currently required to avoid this situation as it is difficult to handle on the server side, especially when the filenames are the same.

packages/core/src/server/prodServer.ts
chenjiahan: Why `output.assetPrefix` is not in `context.environments`, I think these two code are basically the same.

```js
const assetPrefixes  = Object.values(context.environments).map(
  (item) => pathnameParse(item.config.output.assetPrefix),
);
```

```js
const assetPrefixes = Object.values(
  context.normalizedConfig?.environments || {},
).map((e) => pathnameParse(e.output.assetPrefix));
```

packages/core/src/server/compilerDevMiddleware.ts
chenjiahan: get

chenjiahan (approved):

PR #2768, opened 2024-07-03
feat: add environments param into rsbuild api

9aoy:
## Summary

Users can use `environment` info like follows:

```ts
api.onDevCompileDone(async ({ isFirstCompile, stats, environments }) => {
  const entries = Object.keys(environments).map(e => e.entry)
})
```
## Related Links

https://github.com/web-infra-dev/rsbuild/pull/2680

## Checklist

### DIFF
--- a/packages/core/src/server/devServer.ts
+++ b/packages/core/src/server/devServer.ts
@@ -345,6 +345,7 @@ export async function createDevServer<
       await options.context.hooks.onAfterStartDevServer.call({
         port,
         routes,
+        environments: options.context.environments,
       });
     },
     onHTTPUpgrade: devMiddlewares.onUpgrade,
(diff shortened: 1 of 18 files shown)


packages/core/src/server/devServer.ts
```diff
@@ -345,6 +345,7 @@ export async function createDevServer<
       await options.context.hooks.onAfterStartDevServer.call({
         port,
         routes,
+        environments: options.context.environments,
```
chenjiahan, line 348: Should we always put environments at the second parameter? To keep consistent

9aoy: Sounds reasonable, but some hooks don't have a first parameter, which is weird if for consistency...

chenjiahan: get, first param is ok

chenjiahan (approved):

PR #2799, opened 2024-07-04
feat: Check if the current directory is empty.

liangchaofei:
## Checklist

### DIFF
--- a/packages/create-rsbuild/src/index.ts
+++ b/packages/create-rsbuild/src/index.ts
@@ -45,6 +45,22 @@ function isEmptyDir(path: string) {
   return files.length === 0 || (files.length === 1 && files[0] === '.git');
 }
 
+async function checkCwd() {
+  const option = checkCancel<string>(
+    await select({
+      message: 'Current folder is not empty, please choose:',
+      options: [
+        { value: 'yes', label: 'Continue and override files' },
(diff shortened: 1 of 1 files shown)


packages/create-rsbuild/src/index.ts
```diff
@@ -56,6 +72,12 @@ export async function main() {
   const packageJsonPath = path.join(packageRoot, 'package.json');
   const { version } = require(packageJsonPath);
 
+  const currentPath = path.join(cwd);
+  // check cwd is empty
```
chenjiahan, line 76: I'm wondering why do we need to check `cwd` folder

liangchaofei: > I'm wondering why do we need to check `cwd` folder

When I develop scaffolding tools, I usually check if the current directory contains any other files. The scenario I have in mind is that users often want to create a project in a clean directory. This way, we can provide a friendly reminder to the user. Of course, this is not crucial, and you can decide whether to merge this PR.

chenjiahan: We have already checked the `target` folder, and all new files will be output to the `target` folder instead of `cwd`. So I don't understand why we need to check the `cwd` directory.

PR #2824, opened 2024-07-06
refactor(core): use `Promise.all` instead of one by one

colinaaa:
## Summary

Use `Promise.all()` with dynamic imports instead of awaiting one by one.

## Checklist


chenjiahan (approved):
👍

PR #2851, opened 2024-07-09
chore: export some types for use 

9aoy:
## Summary

Export some types for use by high-level frameworks (such as Modern.js).

## Checklist

### DIFF
--- a/packages/core/src/internal.ts
+++ b/packages/core/src/internal.ts
@@ -14,7 +14,6 @@ export {
   getRsbuildInspectConfig,
   outputInspectConfigFiles,
 } from './config';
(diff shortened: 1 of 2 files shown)


packages/core/src/internal.ts
```diff
@@ -14,6 +14,7 @@ export {
   getRsbuildInspectConfig,
   outputInspectConfigFiles,
 } from './config';
+export { getBrowserslist, getBrowserslistByEnvironment } from './createContext';
```
chenjiahan, line 17: We should not expose internal helpers for Modern.js

chenjiahan (approved):

PR #2865, opened 2024-07-10
fix: js/css are minified when both minify.js and minify.css are false

xc2:
## Checklist

### DIFF
--- a/packages/core/src/helpers.ts
+++ b/packages/core/src/helpers.ts
@@ -1,6 +1,11 @@
 import fs from 'node:fs';
 import path, { posix } from 'node:path';
-import type { StatsCompilation, StatsValue } from '@rspack/core';
+import type {
+  LightningCssMinimizerRspackPluginOptions,
+  StatsCompilation,
+  StatsValue,
+  SwcJsMinimizerRspackPluginOptions,
+} from '@rspack/core';
 import deepmerge from 'deepmerge';
 import color from 'picocolors';
 import type RspackChain from 'rspack-chain';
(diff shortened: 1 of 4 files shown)


packages/core/src/helpers.ts
chenjiahan, line 1: Can we add some PR descriptions 😄

packages/core/src/helpers.ts
```diff
@@ -617,3 +622,36 @@ export const prettyTime = (seconds: number): string => {
...
+
+  if (minify === true) {
+    return {
+      minifyJs: true,
+      minifyCss: true,
+    };
+  }
+
+  return {
+    minifyJs: !(minify.js === false || !isProd),
```
chenjiahan, line 652: I think `minify.js` can be enabled in dev mode

chenjiahan: My mistake, forget it..

packages/core/src/plugins/minimize.ts
```diff
@@ -99,22 +63,17 @@ export const pluginMinimize = (): RsbuildPlugin => ({
...
-      const { SwcJsMinimizerRspackPlugin, LightningCssMinimizerRspackPlugin } =
-        rspack;
+      const { minifyJs, minifyCss, jsOptions, cssOptions } = parseMinifyOptions(
+        config,
+        isProd,
+      );
 
-      const { minifyJs, minifyCss, jsOptions, cssOptions } =
-        parseMinifyOptions(config);
+      chain.optimization.minimize(minifyJs || minifyCss);
```
chenjiahan, line 71: `chain.optimization.minimize(minifyJs || minifyCss);` is set by the basic plugin, should we set it here again?

xc2: > `chain.optimization.minimize(minifyJs || minifyCss);` is set by the basic plugin, should we set it here again?

it is for a such edge case that `pluginBasic` is not applied - typically seen in test cases.

actually i'm not sure why it's in `pluginBasic`

chenjiahan: Yeah we can also remove the `chain.optimization.minimize()` from basic plugin.

But this SWC/LightningCSS minimizer plugin is skipped in the webpack mode 😂, so we need to set ``chain.optimization.minimize()` before this `if` condition.

image

xc2: i've refined this PR, please check again.

packages/core/src/plugins/minimize.ts
```diff
@@ -64,25 +64,20 @@ export const parseMinifyOptions = (
...
-    return {
-      minifyJs: false,
-      minifyCss: false,
-    };
+  if (!isProd) {
+    return { minifyJs: false, minifyCss: false };
   }
+  const { minify: _minify } = config.output;
+  const minify =
+    _minify && typeof _minify === 'object'
```
chenjiahan, line 72: I prefer to use the former implementation because it seems simpler. 😂

xc2: Done!

chenjiahan (approved):

PR #2867, opened 2024-07-10
feat: OPTIONS request return success code

vtv6:
## Summary
Support returning success code in OPTIONS requests.
## Checklist

### DIFF
--- a/packages/core/src/server/getDevMiddlewares.ts
+++ b/packages/core/src/server/getDevMiddlewares.ts
@@ -104,6 +104,16 @@ const applyDefaultMiddlewares = async ({
     next();
   });
 
+  middlewares.push((req, res, next) => {
+    if (req.method === 'OPTIONS') {
+      res.statusCode = 200;
+      res.setHeader('Content-Length', '0');
+      res.end();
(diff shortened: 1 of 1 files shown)


chenjiahan:
Hi, can you provide more context for this change?

vtv6:
> Hi, can you provide more context for this change?

vtv6:
I'm integrating with Module Federation v2. Every time the shell app requests the `mf-manifest.json` file, an OPTIONS request is made to check CORS. However, the OPTIONS request always returns a 404 error, causing the CORS check to fail. Therefore, I referred to webpack-dev-server to create this MR

chenjiahan (approved):

chenjiahan:
Thanks for the explanation, I have found the related PR from webpack-dev-server: https://github.com/webpack/webpack-dev-server/pull/4559

I will merge this PR and make some changes:

- Check-pick the commit to main branch, because we have no plan to release 0.x version for new features at present.
- Change the status code from `200` to `204`, because server does not response any data.
- Move the middleware to the last as a fallback.

PR #2905, opened 2024-07-12
docs: port value should be string instead of int

vikas5914:
## Summary

[doc] port value should be string instead of int

## Related Links
From the same doc
```
port?: string;
```

## Checklist


chenjiahan (approved):
thanks!

PR #2919, opened 2024-07-13
refactor(core): use bundler chain to enable Rsdoctor

colinaaa:
## Summary

This makes modifying Rsdoctor options in plugin easier. E.g.:

```js
export function pluginRsdoctor(): RsbuildPlugin {
  return {
    name: 'rsdoctor-modified',
    setup(api) {
      api.modifyBundlerChain((chain, { CHAIN_ID }) => {
        chain
          .plugin(CHAIN_ID.PLUGIN.RSDOCTOR)
          .tap((options: RsdoctorOptions) => modifyRsdoctorOptions(options))
      })
    },
  }
}
```

## Checklist

### DIFF
--- a/packages/core/src/plugins/rsdoctor.ts
+++ b/packages/core/src/plugins/rsdoctor.ts
@@ -11,7 +11,7 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
   name: 'rsbuild:rsdoctor',
 
   setup(api) {
-    api.onBeforeCreateCompiler(async ({ bundlerConfigs }) => {
+    api.modifyBundlerChain(async (chain, { CHAIN_ID }) => {
       if (process.env.RSDOCTOR !== 'true') {
         return;
       }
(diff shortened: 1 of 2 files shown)


packages/core/src/plugins/rsdoctor.ts
```diff
@@ -43,25 +43,13 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
...
-        config.plugins.push(new module[pluginName]());
-        isAutoRegister = true;
+      if (chain.plugins.has(CHAIN_ID.PLUGIN.RSDOCTOR)) {
+        return;
       }
 
-      if (isAutoRegister) {
-        logger.info(`${color.bold(color.yellow(packageName))} enabled.`);
-      }
+      chain.plugin(CHAIN_ID.PLUGIN.RSDOCTOR).use(module[pluginName]);
```
colinaaa, line 50: This may be a breaking change for appending the Rsdoctor plugin like this:

```js
mport { RsdoctorRspackPlugin } from '@rsdoctor/rspack-plugin';

export default {
  tools: {
    rspack(config, { appendPlugins }) {
      if (process.env.RSDOCTOR === 'true') {
        appendPlugins(
          new RsdoctorRspackPlugin({
            // plugin options
          }),
        );
      }
    },
  },
};
```

chenjiahan: This is an unexpected breaking change, we should allow users to define Rsdoctor plugin via `tools.rspack`, that's why Rsbuild uses the `onBeforeCreateCompiler` hook to check Rspack config and enable the Rsdoctor plugin.

colinaaa: That's reasonable. I would find another way to modify the options of Rsdoctor. Thanks!

PR #2989, opened 2024-07-22
fix: webSocket stats message conflict when multiple web environment

9aoy:
## Summary

Each web socket client should only receive stats update messages from its own compilation.

## Checklist


packages/core/src/client/hmr.ts
```diff
@@ -7,6 +7,8 @@
 import type { ClientConfig, StatsError } from '../types';
```
chenjiahan, line 7: It is possible to add an e2e test case for this?

packages/core/src/server/socketServer.ts
```diff
@@ -22,20 +22,34 @@ function isEqualSet(a: Set<string>, b: Set<string>): boolean {
   return true;
 }
 
+const parseUrlParams = (url?: string) => {
```
chenjiahan, line 25: Can we use the Node.js built-in query-string module?

https://nodejs.org/api/querystring.html

packages/core/src/server/socketServer.ts
```diff
@@ -77,32 +80,55 @@ export class SocketServer {
       }
     }, 30000);
 
-    this.wsServer.on('connection', (socket) => {
-      this.onConnect(socket);
+    this.wsServer.on('connection', (socket, req) => {
+      // /rsbuild-hmr?compilationName=web
+      const queryStr = req.url ? req.url?.split('?')[1] : '';
```
chenjiahan, line 85:
```suggestion
      const queryStr = req.url ? req.url.split('?')[1] : '';
```

chenjiahan (approved):

PR #2992, opened 2024-07-23
fix: modify the criteria for verifying the Rsdoctor plugin

easy1090:
## Summary

fix: modify the criteria for verifying the Rsdoctor plugin.

## Checklist

### DIFF
--- a/packages/core/src/plugins/rsdoctor.ts
+++ b/packages/core/src/plugins/rsdoctor.ts
@@ -1,3 +1,4 @@
+import type { Configuration } from '@rspack/core';
 import color from 'picocolors';
 import { logger } from '../logger';
 import type { BundlerPluginInstance, RsbuildPlugin } from '../types';
@@ -7,15 +8,19 @@ type RsdoctorExports = {
   RsdoctorWebpackPlugin: { new (): BundlerPluginInstance };
 };
 
+type MaybeRsdoctorPlugin = Configuration['plugins'] & { isRsdoctorPlugin?: boolean };
+
 export const pluginRsdoctor = (): RsbuildPlugin => ({
   name: 'rsbuild:rsdoctor',
 
   setup(api) {
     api.onBeforeCreateCompiler(async ({ bundlerConfigs }) => {
+      // If Rsdoctor isn't enabled, skip this plugin.
       if (process.env.RSDOCTOR !== 'true') {
         return;
       }
 
+      // Add Rsdoctor plugin to start analysis.
       const isRspack = api.context.bundlerType === 'rspack';
       const packageName = isRspack
         ? '@rsdoctor/rspack-plugin'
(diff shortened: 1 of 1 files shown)


packages/core/src/plugins/rsdoctor.ts
```diff
@@ -45,9 +48,14 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
 
       let isAutoRegister = false;
 
+      function isRsdoctorPlugin(plugin: Configuration['plugins'] & { isRsdoctorPlugin?: boolean }) {
+        return plugin != null && typeof plugin.isRsdoctorPlugin === 'boolean';
```
chenjiahan, line 52: `!=` is not recommended to use

```suggestion
        return plugin?.isRsdoctorPlugin === true;
```

packages/core/src/plugins/rsdoctor.ts
```diff
@@ -45,9 +48,14 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
 
       let isAutoRegister = false;
 
+      function isRsdoctorPlugin(plugin: Configuration['plugins'] & { isRsdoctorPlugin?: boolean }) {
```
chenjiahan, line 51: function -> const + arrow function

packages/core/src/plugins/rsdoctor.ts
```diff
@@ -45,9 +48,14 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
...
+        return plugin != null && typeof plugin.isRsdoctorPlugin === 'boolean';
+      }
+
       for (const config of bundlerConfigs) {
-        const registered = config.plugins?.some(
-          (plugin) => plugin?.constructor?.name === pluginName,
+        // If user has add Rsdoctor plugin in config file, will return.
```
chenjiahan, line 56:
```suggestion
        // If user has added the Rsdoctor plugin to the config file, it will return.
```

packages/core/src/plugins/rsdoctor.ts
```diff
@@ -12,10 +13,12 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
 
   setup(api) {
     api.onBeforeCreateCompiler(async ({ bundlerConfigs }) => {
+      // If don't start RSDOCTOR analysis will return.
```
chenjiahan, line 16:
```suggestion
      // If Rsdoctor isn't enabled, skip this plugin.
```

packages/core/src/plugins/rsdoctor.ts
```diff
@@ -45,9 +48,14 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
...
+      function isRsdoctorPlugin(plugin: Configuration['plugins'] & { isRsdoctorPlugin?: boolean }) {
+        return plugin != null && typeof plugin.isRsdoctorPlugin === 'boolean';
+      }
+
       for (const config of bundlerConfigs) {
-        const registered = config.plugins?.some(
-          (plugin) => plugin?.constructor?.name === pluginName,
+        // If user has add Rsdoctor plugin in config file, will return.
+        const registered = config.plugins?.some(          
+          (plugin) => isRsdoctorPlugin(plugin as unknown as Configuration['plugins'] & { isRsdoctorPlugin?: boolean }) || plugin?.constructor?.name === pluginName,
```
chenjiahan, line 58: We can move the `plugin?.constructor?.name === pluginName` to `isRsdoctorPlugin()` helper.

chenjiahan, line 58: We can declare a type to reuse it

```suggestion
type MaybeRsdoctorPlugin = Configuration['plugins'] & { isRsdoctorPlugin?: boolean })
```

easy1090: > We can move the `plugin?.constructor?.name === pluginName` to `isRsdoctorPlugin()` helper.

@chenjiahan I just worry about remove the `plugin?.constructor?.name === pluginName` may be a break change.

chenjiahan: No, I mean move the logic to the function, not to remove it.

chenjiahan (approved):
👍

PR #2986, opened 2024-07-22
feat: support add plugins for specified environment

9aoy:
## Summary

Support add plugins for specified environment.

For example, enable the react plugin only in the web environment:

```ts title=rsbuild.config.ts
import { pluginReact } from '@rsbuild/plugin-react';

export default {
  environments: {
    web: {
      output: {
        target: 'web',
      },
      plugins: [pluginReact()],
    },
    node: {
      output: {
        target: 'node',
      },
    },
  },
};
```

TODO: Some plugins do not yet support single-environment registration because they use `api.modifyRsbuildConfig` hook to modify the global configuration.

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/2620

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #2994, opened 2024-07-23
feat: more plugins support used as environment plugin

9aoy:
## Summary

use `modifyEnvironmentConfig` instead of `modifyRsbuildConfig`, make more plugins support used as environment plugin.

## Related Links

https://github.com/web-infra-dev/rsbuild/pull/2986

## Checklist

### DIFF
--- a/packages/plugin-source-build/src/plugin.ts
+++ b/packages/plugin-source-build/src/plugin.ts
@@ -58,15 +58,17 @@ export function pluginSourceBuild(
     setup(api) {
       const projectRootPath = api.context.rootPath;
 
-      let projects: Project[] = [];
+      let projects: Project[] | undefined;
 
-      api.modifyRsbuildConfig(async (config) => {
-        projects = await getDependentProjects(projectName || projectRootPath, {
-          cwd: projectRootPath,
-          recursive: true,
-          filter: filterByField(sourceField, true),
-          extraMonorepoStrategies,
-        });
+      api.modifyEnvironmentConfig(async (config) => {
+        projects =
+          projects ||
+          (await getDependentProjects(projectName || projectRootPath, {
+            cwd: projectRootPath,
+            recursive: true,
+            filter: filterByField(sourceField, true),
+            extraMonorepoStrategies,
+          }));
 
         const includes = await getSourceInclude({
           projects,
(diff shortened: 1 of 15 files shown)


packages/plugin-source-build/src/plugin.ts
```diff
@@ -58,15 +58,17 @@ export function pluginSourceBuild(
...
-          cwd: projectRootPath,
-          recursive: true,
-          filter: filterByField(sourceField, true),
-          extraMonorepoStrategies,
-        });
+      let projects: Project[] | undefined;
+
+      api.modifyEnvironmentConfig(async (config) => {
+        projects =
+          projects ||
```
chenjiahan, line 65: What if each environments use the different projects name?

If we want to cache `project` in the global, we may need to create a `Map` and use `projectName || projectRootPath` as the cache key.

packages/plugin-styled-components/src/index.ts
```diff
@@ -57,39 +59,43 @@ export const pluginStyledComponents = (
       });
     };
 
+    // use modifyRsbuildConfig to get useSSR default value (can be override by pluginOptions)
```
chenjiahan, line 62: From my perspective, we should not use the `modifyRsbuildConfig` to `get` config.

If we want to get the config, we can use `api.getRsbuildConfig()`.

packages/plugin-source-build/src/plugin.ts
9aoy: This is just to ensure the performance between different Environments of the global plugin. `different projects name` will only appear when plugins are registered separately.

chenjiahan: Ok I get it

packages/plugin-styled-components/src/index.ts
9aoy: Calling `api.getRsbuildConfig()` in modifyEnvironmentConfig hook is currently not supported because RsbuildConfig contains EnvironmentConfig, which has not yet been determined.

chenjiahan: No, it is allowed call `api.getRsbuildConfig()` anywhere and it should alway return the current (latest) config.

For example, if `modifyRsbuildConfig` has been called, `api.getRsbuildConfig()` should return the normalized Rsbuild config.

image

9aoy: get it!

chenjiahan (approved):

Issue #2957, opened 2024-07-18
[Bug]: `output.inlineStyles: true` doesn't seem to work in dev


matteom-synth:
### Version

```sh
System:
    OS: macOS 14.5
    CPU: (11) arm64 Apple M3 Pro
    Memory: 554.95 MB / 36.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Browsers:
    Chrome: 126.0.6478.182
    Safari: 17.5
  npmPackages:
    @rsbuild/core: 1.0.1-beta.3 => 1.0.1-beta.3
    @rsbuild/plugin-react: 1.0.1-beta.3 => 1.0.1-beta.3
```

### Details

I'm not sure if I have some wrong expectations, but [`output.inlineStyles: true`](https://rsbuild.dev/config/output/inline-styles) doesn't seem to inline stylesheets into the resulting HTML in dev. Is it supposed to work only on production builds?

### Reproduce link

https://github.com/matteom-synth/rsbuild-issue-repro-3

### Reproduce Steps

1. Clone the repo.
2. Run `pnpm install`.
3. Run `pnpm dev`.
4. Check the HTML output. My expectation is that, given that [`output.inlineStyles` is set to `true`](https://github.com/matteom-synth/rsbuild-issue-repro-3/blob/f42d5b426a9ced04ec950924e68a576140205663/rsbuild.config.ts#L7), the stylesheet should be inlined into the HTML. But it doesn't seem to be the case: 
<img width="1104" alt="image" src="https://github.com/user-attachments/assets/f6d73648-943c-450e-8e21-5202578d7832">

matteom-synth:
After a quick investigation, I guess it's because of this check, right? https://github.com/web-infra-dev/rsbuild/blob/b46f3af10ca198081982c318865eebc2b4407cfd/packages/core/src/plugins/inlineChunk.ts#L11

Any reason why this is being disabled in `dev`? I guess it might be because it works better with HMR? Asking because sadly this is breaking our dev env (for a specific edge case) 🙃 because in our Webpack config by default styles are inlined in dev as well, and this is causing a differernt behaviour.

Thulof:
> After a quick investigation, I guess it's because of this check, right?
> 
> https://github.com/web-infra-dev/rsbuild/blob/b46f3af10ca198081982c318865eebc2b4407cfd/packages/core/src/plugins/inlineChunk.ts#L11
> 
[...]

PR #3000, opened 2024-07-23
feat: enable inline chunks based on conditions through output options

Thulof:
## Summary
Support inline chunks in development mode

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/2957

## Checklist

### DIFF
--- a/e2e/cases/inline-chunk/index.test.ts
+++ b/e2e/cases/inline-chunk/index.test.ts
@@ -1,5 +1,5 @@
 import path from 'node:path';
-import { build, gotoPage } from '@e2e/helper';
+import { build, dev, gotoPage } from '@e2e/helper';
 import { expect, test } from '@playwright/test';
 import type { RspackChain } from '@rsbuild/core';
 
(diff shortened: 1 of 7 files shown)


chenjiahan:
Thanks for your contribution! 

I prefer extend the current `output.inlineStyles` config instead of adding a new `dev.inlineStyles` config.

We can extend the config type like this:

```ts
type InlineStylesTest =
  | RegExp
  | ((params: { size: number; name: string }) => boolean);

type InlineStyles =
  | boolean
  | InlineStylesTest
  | {
      enable?: boolean | 'auto';
      text: InlineStylesTest;
    };
```

Example:

```ts
export default {
  output: {
    inlineStyles: {
      enable: true,
      test: /[\\/]main\.\w+\.css$/,
    },
  },
};
```

And `output.inlineStyles: true` is equivalent to `enable: 'auto'`, which means it is only enabled in production build.

```ts
export default {
  output: {
    inlineStyles: {
      enable: 'auto',
    },
  },
};
```

If `enable` is `true`, it will works in both dev build and prod build:

```ts
export default {
  output: {
    inlineStyles: {
      enable: true,
    },
  },
};
```

chenjiahan (approved):
Awesome! Thanks ❤️

PR #3028, opened 2024-07-26
feat(core): add the `--watch-files` arg to the devCommand

kurorinto:
## Summary
support `devCommand` specify the configuration files to listen to restart the dev server.

## Related Links
https://github.com/web-infra-dev/rsbuild/discussions/3021

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #3057, opened 2024-07-29
feat(core): analyzing the modules that rsbuild.config.ts depends on and watching them.

kurorinto:
## Summary
support analysis of `rsbuild.config.ts`'s dependencies and restart devServer when they change.
## Related Links
https://github.com/web-infra-dev/rsbuild/pull/3028

## Checklist


packages/core/src/cli/analyzingDependencies.ts
```diff
@@ -0,0 +1,125 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import * as acorn from 'acorn';
+import { tsPlugin } from 'acorn-typescript';
```
chenjiahan, line 4: `acorn` is too slow for Rsbuild, and we do not want to introduce extra dependencies for this 😂

kurorinto: I later saw that `vite` is using `esbuild.build` to get the dependency of the config.[(link)](https://github.com/vitejs/vite/blob/b947fdcc9d0db51ee6ac64d9712e8f04077280a7/packages/vite/src/node/config.ts#L1214)
Perhaps config file dependency analysis should be done at rspack?

chenjiahan: I think it is possible to replace `jiti` with Rspack and get the dependency graph from Rspack. 

But there are two concerns:

- If the extension of the config file is not `.ts`, Rsbuild will not build or transform the config file (to avoid the transformation overhead).
- If the extension is `.ts`, using `jiti` can cover more edge cases because it is designed to load config and can handle interoperability between ESM and CommonJS.

kurorinto: Do you mean that `jiti` is used to ensure the interoperability between esm and cjs, so there is no good way to resolve module dependencies?

chenjiahan: It's more of a trade-off, we need to balance startup performance (loading config faster) and development experience (watching dependent modules of config file).

kurorinto: Is it possible to make a dev option in `rsbuild.config.ts`? Because sometimes users need to change config frequently, like I need to switch `server.proxy` frequently during development, it was a bit of a hassle to go to the terminal on another screen and restart `npm run dev` after I changed the `server.proxy`.Also, thanks to the rsbuild team for a great developer experience!

chenjiahan: Thanks ❤️ I think we can add a `type` option to [dev.watchFiles](https://rsbuild.dev/zh/config/dev/watch-files):

- `type: 'reload-page'` is the default value, align with the current usage.

```js
export default {
  dev: {
    watchFiles: {
      type: 'reload-page',
      paths: ['public/foo.txt'],
    },
  },
};
```


- `type: 'reload-server'` is the new value, can be used to watch some config files and reload the server.

```js
export default {
  dev: {
    watchFiles: {
      type: 'reload-server',
      paths: ['configs/rsbuild.base.config.ts'],
    },
  },
};
```

How do you think about this API? Do you want to implement it 😄

kurorinto: I'll try right away.

Issue #3094, opened 2024-08-01
[Feature]: Support `<port>` placeholder in `dev.assetPrefix`


colinaaa:
### What problem does this feature solve?

We want to specify a custom `dev.assetPrefix` with hostname and port. But when the port is occupied, Rsbuild will automatically bump up the port by default, making the `dev.assetPrefix` incorrect.

It would be great if the `dev.assetPrefix` supports using the `<port>` placeholder.

BTW, I'm not sure if `output.assetPrefix` needs this.

### What does the proposed API look like?

The same as `dev.client.port` and `server.open`:

```js
// rsbuild.config.js
export default {
  dev: {
    assetPrefix: 'http://hostname:<port>/',
    client: {
      hostname: 'hostname',
      port: '<port>',
    },
  },
  server: {
    open: 'http://hostname:<port>/',
  },
}

chenjiahan:
PR welcome 😄

PR #3098, opened 2024-08-01
feat: support using `<port>` in `dev.assetPrefix`

colinaaa:
## Summary

fix: #3094 

Note that `<port>` is also available in `output.assetPrefix`. But I can't think of any situation in which it should be used.

## Checklist


chenjiahan (approved):

chenjiahan:
Nice! 👍

PR #3189, opened 2024-08-12
feat!: output inspect config to .rsbuild dir

9aoy:
## Summary

output inspect config to `.rsbuild` directory by default, and we will output more rsbuild related files to the `.rsbuild` directory.

image

## Checklist

### DIFF
--- a/packages/core/src/plugins/cleanOutput.ts
+++ b/packages/core/src/plugins/cleanOutput.ts
@@ -1,5 +1,6 @@
-import { sep } from 'node:path';
+import { join, sep } from 'node:path';
 import color from 'picocolors';
+import { RSBUILD_OUTPUTS_PATH } from '../constants';
 import { emptyDir } from '../helpers';
 import { logger } from '../logger';
 import type { EnvironmentContext, RsbuildPlugin } from '../types';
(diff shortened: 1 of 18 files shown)


packages/core/src/plugins/cleanOutput.ts
```diff
@@ -16,6 +17,23 @@ export const pluginCleanOutput = (): RsbuildPlugin => ({
...
+    // clean rsbuild outputs, such as inspect files
+    const cleanRsbuildOutputs = async () => {
+      const { rootPath, distPath } = api.context;
+      const config = api.getNormalizedConfig();
+      const cleanPath = join(distPath, RSBUILD_OUTPUTS_PATH);
+
+      let { cleanDistPath } = config.output;
+
+      if (cleanDistPath === undefined) {
+        cleanDistPath = isStrictSubdir(rootPath, cleanPath);
```
chenjiahan, line 29: We can avoid calling `isStrictSubdir` multiple times

packages/core/src/plugins/cleanOutput.ts
```diff
@@ -16,6 +17,23 @@ export const pluginCleanOutput = (): RsbuildPlugin => ({
...
+      }
+
+      if (cleanDistPath) {
+        await emptyDir(cleanPath);
```
chenjiahan, line 33: `emptyDir` is slow, can we find a method to avoid calling it multiple times?

chenjiahan (approved):

PR #3196, opened 2024-08-13
fix: onBeforeBuild hook should before onBeforeEnvironmentCompile hook

9aoy:
## Summary

fix build hooks order, `onBeforeBuild` hook should before all `onBeforeEnvironmentCompile` hooks run.

## Checklist


packages/core/src/hooks.ts
```diff
@@ -230,15 +232,16 @@ const onBeforeCompile = ({
...
-          // ensure only last compiler done will trigger beforeCompile hook
-          // avoid other compiler done triggers when executing async beforeEnvironmentCompiler hook
-          const lastCompilerDone = doneCompilers === compilers.length;
-
-          await beforeEnvironmentCompiler(index);
+          const isFirstTrigger = doneCompilers === 1;
 
-          if (lastCompilerDone) {
-            await beforeCompile?.();
+          if (isFirstTrigger) {
```
chenjiahan, line 237: It seems we do not need a `isFirstTrigger` variable:

```suggestion
          if (!waitBeforeCompileDone) {
```

chenjiahan: `doneCompilers` is not reliable as it can decrease in some cases.

chenjiahan (approved):

Issue #1999, opened 2024-04-04
[Bug]: SCSS files not loaded


JoseBuendiaDigio:
### Version

```sh
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
    Memory: 11.40 GB / 15.48 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  npmPackages:
    @rsbuild/core: 0.5.2 => 0.5.2 
    @rsbuild/plugin-eslint: ^0.5.2 => 0.5.2 
    @rsbuild/plugin-react: 0.5.2 => 0.5.2
```

### Details

I'm currently migrating from RsPack to RsBuild but SCSS seems to not be loaded.

Here is how I config it in RsPack

```
{
  test: /\.scss$/,
  use: [
    'style-loader',
    {
      loader: 'css-loader',
      options: {
        importLoaders: 3,
        sourceMap: false,
        modules: {
          mode: 'icss',
        },
      },
    },
    {
      loader: 'postcss-loader',
      options: {
        postcssOptions: {
          ident: 'postcss',
          config: false,
          sourceMap: false,
          
        },
      },
    },
    {
      loader: 'resolve-url-loader',
    },
    {
      loader: 'sass-loader',
      options: {
        sourceMap: true,
        
      },
    },
  ],
}
```

But in RsBuild I tried some things with styleLoader and postscss but is not loading scss files (it load it as undefined).

How should it load it?

### Reproduce link

https://github.com/JoseBuendiaDigio/rsbuild-scss

### Reproduce Steps

rsbuild build and rsbuild dev

JoseBuendiaDigio:
Maybe I need to add css-loader and style-loader to scss rule but I don't know how to add it. I tried things like this, but without result.

`config.module.rules[10].oneOf[1].use.unshift(styleLoader, cssLoader);`

chenjiahan:
Rsbuild provides builtin supports for SCSS, so user do not need to add the css-loader/scss-loader/postcss-loader.

If you want to use `style-loader` in Rsbuild, you can enable this option: https://rsbuild.dev/config/output/inject-styles

JoseBuendiaDigio:
Yes, I tried activating this option but then fails the build becauses it can read the scss values when readed from js

JoseBuendiaDigio:
[...]

PR #3199, opened 2024-08-13
feat(core): Remove the default exportOnlyLocals value from cssloader …

witsaint:
## Summary
Remove the default value of exportOnlyLocals when cssmodule is enabled，

When module is enabled, exportOnlyLocals is set to true by default，the exported value is incorrectly mounted

image

Therefore, exportOnlyLocals does not need a default value, leaving it up to the user to select

## Related Links
[[Bug]: SCSS files not loaded #1999](https://github.com/web-infra-dev/rsbuild/issues/1999)

## Checklist

### DIFF
--- a/packages/core/src/plugins/css.ts
+++ b/packages/core/src/plugins/css.ts
@@ -35,40 +35,6 @@ const getCSSModulesLocalIdentName = (
     ? '[local]-[hash:base64:6]'
     : '[path][name]__[local]-[hash:base64:6]');
 
-// If the target is not `web` and the modules option of css-loader is enabled,
-// we must enable exportOnlyLocals to only exports the modules identifier mappings.
-// Otherwise, the compiled CSS code may contain invalid code, such as `new URL`.
-// https://github.com/webpack-contrib/css-loader#exportonlylocals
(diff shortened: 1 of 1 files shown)


chenjiahan:
Does https://github.com/web-infra-dev/rsbuild/issues/1999 related to `exportOnlyLocals`?

Rsbuild only sets `exportOnlyLocals` to `true` if `output.target` is not `'web'`, but #1999  does not seem to set `output.target.

witsaint:
> Does #1999 related to `exportOnlyLocals`?
> 
> Rsbuild only sets `exportOnlyLocals` to `true` if `output.target` is not `'web'`, but #1999 does not seem to set `output.target`.

Yes ,I set `false`, compile fine
image

image

chenjiahan:
Just run `DEBUG=rsbuild pnpm build` and you will find that Rsbuild does not set `exportOnlyLocals` by default.

image

witsaint:
OK, It seems that the error report is not stable. There are other errors reported here. I will close MR First and then continue to see

witsaint:
It seems that the error is not stable. There are other errors reported here. I will close MR First

PR #3216, opened 2024-08-15
feat!: use `connectWebSocket` instead of `onHTTPUpdate` method

9aoy:
## Summary

This PR will cause **breaking change** for users who use Rsbuild's `createDevServer` JavaScript API directly.

The `﻿onHTTPUpgrade` method does not fully convey its actual purpose, which is to activate the webSocket connection for Rsbuild.

Users who use rsbuild createDevServer directly need to make the following modifications:
```diff
  const rsbuildServer = await rsbuild.createDevServer();

  const httpServer = app.listen(rsbuildServer.port);

-  httpServer.on('upgrade', rsbuildServer.onHTTPUpgrade);
+  rsbuildServer.connectWebSocket({ server: httpServer });

```

## Checklist

### DIFF
--- a/e2e/cases/server/custom-server/scripts/pureServer.mjs
+++ b/e2e/cases/server/custom-server/scripts/pureServer.mjs
@@ -48,9 +48,6 @@ export async function startDevServerPure(fixtures) {
     await rsbuildServer.afterListen();
   });
 
-  // subscribe the server's http upgrade event to handle WebSocket upgrade
-  server.on('upgrade', rsbuildServer.onHTTPUpgrade);
-
   return {
     config: { port: rsbuildServer.port },
     close: async () => {
--- a/website/docs/en/api/javascript-api/instance.mdx
+++ b/website/docs/en/api/javascript-api/instance.mdx
@@ -360,6 +360,27 @@ const server = await rsbuild.createDevServer();
 await server.listen();
 ```
 
+### connectSocket
+
+Rsbuild has a builtin Socket server to support HMR:
+
+1. When a user accesses a page through browser, a Web Socket connection request is automatically initiated to the server.
+2. After the Rsbuild dev server detects the connection request, it instructs the builtin Socket server to process it.
+3. After the browser successfully establishes a connection with the Rsbuild Socket server, real-time communication is possible.
+4. The Rsbuild Socket server notifies the browser after each recompilation is complete. The browser then sends a `hot-update.(js|json)` request to the dev server to load the new compiled module.
+
+When you use custom server, you may encounter HMR connection error problems. This is because the custom server does not forward Web Socket connection requests to Rsbuild's Socket server.
+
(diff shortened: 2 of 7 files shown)


e2e/cases/server/custom-server/scripts/pureServer.mjs
```diff
@@ -48,9 +48,6 @@ export async function startDevServerPure(fixtures) {
     await rsbuildServer.afterListen();
   });
 
-  // subscribe the server's http upgrade event to handle WebSocket upgrade
-  server.on('upgrade', rsbuildServer.onHTTPUpgrade);
```
chenjiahan, line 52: should we call `rsbuildServer.connectSocket()` instead?

website/docs/en/api/javascript-api/instance.mdx
```diff
@@ -347,8 +346,7 @@ export async function startDevServer() {
     await rsbuildServer.afterListen();
   });
 
-  // Subscribe to the server's http upgrade event to handle WebSocket upgrades
-  httpServer.on('upgrade', rsbuildServer.onHTTPUpgrade);
+  rsbuildServer.connectSocket({ server: httpServer });
```
chenjiahan, line 349: We need more documentation for this API than just one line in the example.

e2e/cases/server/custom-server/scripts/pureServer.mjs
9aoy: The `connectSocket` and `onHTTPUpgrade` methods are not needed in this case because compilation is not triggered here.

website/docs/en/api/javascript-api/instance.mdx
```diff
@@ -362,6 +360,27 @@ const server = await rsbuild.createDevServer();
...
+### connectSocket
+
+Rsbuild has a builtin Socket server to support HMR:
+
+1. When a user accesses a page through browser, a Web Socket connection request is automatically initiated to the server.
+2. After the Rsbuild dev server detects the connection request, it instructs the builtin Socket server to process it.
+3. After the browser successfully establishes a connection with the Rsbuild Socket server, real-time communication is possible.
+4. The Rsbuild Socket server notifies the browser after each recompilation is complete. The browser then sends a `hot-update.(js|json)` request to the dev server to load the new compiled module.
+
+When you use custom server, you may encounter HMR connection error problems. This is because the custom server does not forward Web Socket connection requests to Rsbuild's Socket server.
```
chenjiahan, line 372: Is it a `Socket server` or a `Socket middleware`? Rsbuild does not start server when using custom server.

9aoy: how about `Socket handler`?Users will misunderstand socket middleware and other middlewares (why `app.use(rsbuildServer.middlewares)` is useless)

chenjiahan: LGTM

chenjiahan (approved):

Issue #3247, opened 2024-08-19
[Bug]: rsbuild injectStyles: true and chunkSplit: all-in-one will inject lots \n and blank space with cssnano useless


sbyps:
### Version

```sh
System:
    OS: macOS 14.6.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 50.20 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 127.0.6533.120
    Safari: 17.6
  npmPackages:
    @rsbuild/core: ^1.0.1-beta.14 => 1.0.1-beta.14 
    @rsbuild/plugin-check-syntax: ^1.0.1 => 1.0.1 
    @rsbuild/plugin-sass: ^1.0.1-beta.14 => 1.0.1-beta.14 
    @rsbuild/plugin-vue2: ^1.0.1-beta.14 => 1.0.1-beta.14
```

### Details

following the official docs: https://rsbuild.dev/zh/config/output/inject-styles#%E4%BD%BF%E7%94%A8%E5%9C%BA%E6%99%AF
then set config like
injectStyles: true,
chunkSplit: all-in-one

will inject lots \n and blank space with cssnano useless in js files

### Reproduce link

https://github.com/sbyps/rspack-demo

### Reproduce Steps

npm run build

cat the js files, search "\n ", will find lots of it
<img width="1523" alt="image" src="https://github.com/user-attachments/assets/bc19368a-f47a-4fc4-af3d-84543e3232b6">

sbyps:
@chenjiahan this PR resolves some case with \n; but we can found lots of \n in sourcemap and js files;
rsbuild version: 1.0.1-rc.4
https://github.com/sbyps/rspack-demo
<img width="911" alt="image" src="https://github.com/user-attachments/assets/1b3d2875-21c7-4b5a-94ba-73e3b03c8479">
<img width="919" alt="image" src="https://github.com/user-attachments/assets/86c3a497-9589-4b2e-bb5c-67c5e22441fb">

chenjiahan:
This is as expected, the source map contains the source code

sbyps:
> This is as expected, the source map contains the source code

![image](https://github.com/user-attachments/assets/dd9b57ec-4799-4ae5-be45-ee0bc031318b)
this image is not sourcemap but production file, also has \n

chenjiahan:
Can you provide a reproduction?

sbyps:
[...]

PR #3254, opened 2024-08-20
fix(core): Disable buildin lightingcss when inject style

witsaint:
## Summary

For production builds, it is recommended to use the default behavior of Rsbuild, which extracts CSS into separate bundles to allow browsers to load CSS and JS assets in parallel.

## Related Links

[[Bug]: rsbuild injectStyles](https://github.com/web-infra-dev/rsbuild/issues/3247)

## Checklist

### DIFF
--- a/packages/core/src/plugins/css.ts
+++ b/packages/core/src/plugins/css.ts
@@ -237,10 +237,11 @@ async function applyCSSRule({
   rule.use(CHAIN_ID.USE.CSS).loader(getCompiledPath('css-loader'));
 
   if (target === 'web') {
-    // `builtin:lightningcss-loader` is not supported when using webpack
+    // `builtin:lightningcss-loader` is not supported when using webpack and inject styles
     if (
       context.bundlerType === 'rspack' &&
-      config.tools.lightningcssLoader !== false
+      config.tools.lightningcssLoader !== false &&
+      config.output.injectStyles !== true
     ) {
       importLoaders++;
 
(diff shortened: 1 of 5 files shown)


chenjiahan (commented):
Can you explain why do we need to disable builtin lightingcss when using `injectStyles`?

witsaint:
> Can you explain why do we need to disable builtin lightingcss when using `injectStyles`?

1: As described in our documentation, when we open injectStyles,If you need to enable this option in production mode, please note that the inlined CSS code will not go through Rsbuild's default CSS minimizer.

2: 
a： When I debugger Rsbuild，I found that there were five loaders before stylePostLoader
image

b：The latest loader is builtin lightingcss
image
The same is shown in rspack.config.web.mjs

c：I found that after using postcss with cssnano, source works fine，But by buildin lightingcss,Found a lot of \n
image

d: When I remove buildin lightingcss in loader, css compression is fine when injectStyles is turned on

chenjiahan:
Get. I believe this fix is not the optimal solution as it prevents PostCSS and LightningCSS from working together.

We can solve this problem by simply adding a `minify` option to Rspack's lightningcss-loader and enanbling minify when `output.injectStyles` is used.

chenjiahan:
Sorry, we should not change the order of lightningcss-loader and postcss-loader, this is an unexpected change.

Take Tailwind CSS as an example. Tailwind CSS it is a PostCSS plugin, it will generate some CSS code, then Lightning CSS will add vendor prefixes to those CSS properties. If we change the loader order, this will be broken.

chenjiahan:
If we need to adjust the default loader behavior, it is best to first discuss the impact of doing so.

PR #3264, opened 2024-08-22
feat: add `environments` to `onAfterStartProdServer`

colinaaa:
## Summary

I have a Rsbuild plugin to print QRCode for all entries. It works well with `rsbuild dev`:

```js
      api.onAfterStartDevServer(async ({ environments }) => {
        const fooEnvironment = environments['foo']
        const entries = Object.keys(fooEnvironment.entry)
        await printQRCode(entries)
      })
```

I would also like to print the similar QRCode for `rsbuild preview`. But I cannot get the entries from `onAfterStartProdServer`.

This patch simply add `environments` to `onAfterStartProdServer`, which will work the same as `onAfterStartDevServer`.

## Related Links

N/A

## Checklist

### DIFF
--- a/packages/core/src/server/prodServer.ts
+++ b/packages/core/src/server/prodServer.ts
@@ -193,6 +193,7 @@ export async function startProdServer(
         await context.hooks.onAfterStartProdServer.call({
           port,
           routes,
+          environments: context.environments,
         });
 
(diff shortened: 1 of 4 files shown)


chenjiahan (approved):
LGTM  👍

PR #3266, opened 2024-08-22
feat: transformImport support function type

9aoy:
## Summary

support modify previous `transformImport` value via  `transformImport`  function.

eg:
```ts
export default {
  source: {
    transformImport: (imports) => {
      return imports.filter(import => import.libraryName !== 'antd');
    },
  },
};
```

## Checklist

### DIFF
--- a/packages/compat/plugin-swc/src/utils.ts
+++ b/packages/compat/plugin-swc/src/utils.ts
@@ -26,9 +26,6 @@ const castArray = <T>(arr?: T | T[]): T[] => {
   return Array.isArray(arr) ? arr : [arr];
 };
 
-const isFunction = (func: unknown): func is (...args: any[]) => any =>
-  typeof func === 'function';
-
 async function findUp({
   filename,
   cwd = process.cwd(),
@@ -205,7 +202,7 @@ const reduceTransformImportConfig = (
 
   let imports: TransformImport[] = [];
   for (const item of castArray(options)) {
-    if (isFunction(item)) {
+    if (typeof item === 'function') {
       imports = item(imports) ?? imports;
     } else {
       imports.push(item);
--- a/packages/compat/plugin-swc/src/utils.ts
+++ b/packages/compat/plugin-swc/src/utils.ts
@@ -193,7 +193,23 @@ const isPrimitiveTransformImport = (
     (item) => Object.prototype.toString.call(item) === '[object Object]',
   );
 
-const applyTransformImportChain = (
+const reduceTransformImportConfig = (
+  options: NormalizedSourceConfig['transformImport'],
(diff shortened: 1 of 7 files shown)


packages/compat/plugin-swc/src/utils.ts
```diff
@@ -184,6 +186,34 @@ const getCoreJsVersion = (corejsPkgPath: string) => {
...
 };
 
+const isPrimitiveTransformImport = (
+  items: unknown[],
+): items is Array<TransformImport> =>
+  items.every(
+    (item) => Object.prototype.toString.call(item) === '[object Object]',
+  );
+
+const applyTransformImportChain = (
```
chenjiahan, line 196: - No need to do recursive calls after correct the types
- No need to pass `default` value, always defaults to `[]`

```suggestion
const reduceTransformImportConfig = (
  options: NormalizedSourceConfig['transformImport'],
): TransformImport[] | false => {
  if (options === false || !options) {
    return false;
  }

  let imports: TransformImport[] = [];
  for (const item of castArray(options)) {
    if (isFunction(item)) {
      imports = item(imports) ?? imports;
    } else {
      imports.push(item);
    }
  }
  return imports;
};
```

packages/core/src/types/config/source.ts
```diff
@@ -61,7 +61,7 @@ export interface SourceConfig {
   /**
    * Used to import the code and style of the component library on demand.
    */
-  transformImport?: false | TransformImport[];
+  transformImport?: ConfigChain<TransformImport[]> | false;
```
chenjiahan, line 64: We can not use `ConfigChain` here, because `TransformImport[][]` is unexpected.

```suggestion
type TransformImportFn = (
  imports: TransformImport[],
) => TransformImport[] | void;

  transformImport?: 
    | TransformImportFn
    | Array<TransformImport | TransformImportFn>
    | false;;
```

packages/compat/plugin-swc/src/utils.ts
```diff
@@ -184,6 +196,24 @@ const getCoreJsVersion = (corejsPkgPath: string) => {
...
+const reduceTransformImportConfig = (
+  options: NormalizedSourceConfig['transformImport'],
+): TransformImport[] | false => {
+  if (options === false || !options) {
+    return false;
+  }
+
+  let imports: TransformImport[] = [];
+  for (const item of castArray(options)) {
+    if (isFunction(item)) {
```
chenjiahan, line 208: Seems no need to create a `isFunction` helper here:

```suggestion
    if (typeof item === 'function') {
```

chenjiahan (approved):

PR #3322, opened 2024-08-30
doc: improve security sri.mdx

liyincode:
## Summary
I think the example is not very strict. Because after configuring sri, Rsbuild can only calculate the integrity value for locally compiled assets, and the url of the third party cdn can only be filled in by the user in the `template` or `html.tags`.     
## Checklist


chenjiahan (commented):
How about add `output.assetPrefix: 'https://cdn.com/'` to the example Rsbuild config?

chenjiahan:
SRI is usually used to verify CDN resources

liyincode:
If the current project is packaged with cdn content, I think the `output.assetPrefix: 'https://cdn.com/'` is appropriate. If the current project is a normal web project, I think the example will cause problems, why the resource link with url prefix will be generated

chenjiahan:
This is just an example to show a common usage, how could it cause any problems?

liyincode:
Just a little confusing for users, I suggest that the document can highlight the usage scenarios better, such as:

1. SRI for locally compiled resources
2. SRI for CDN resources

There will be two examples to tell users how to configure, for example, in the second scenario, you can set `output.assetPrefix: 'https://cdn.com/'`.

chenjiahan:
It is unnecessary to provide two examples, it is too verbose, and MDN also uses the use of CDN as an example: https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/SRI#examples
image

If you feel that `output.assetPrefix` is confusing, we can add a comment to indicate that it is optional and is only used for example.

liyincode:
Ok, I see what you mean. Just keep it the way it is. It makes the most sense

PR #3369, opened 2024-09-03
chore: format all files

chenjiahan:
## Summary

Use prettier to format all files.

## Checklist


fi3ework:
It's just curious about why there's no use of Biome to format all files, especially some files are just normal TS files. 🤔

chenjiahan:
See https://github.com/web-infra-dev/rsbuild/pull/2669

We can mix Biome and Prettier, but I haven't seen much benefit.

Issue #3358, opened 2024-09-03
[Bug]: When starting with Rsdoctor, I still get the message "rspack-plugin" even though it has been installed.


zhangdaren:
### Version

```sh
windows
edge
 "@rsbuild/core": "^1.0.1-rc.0",
    "@rsbuild/plugin-babel": "^1.0.1-rc.0",
    "@rsbuild/plugin-less": "^1.0.1-rc.0",
    "@rsbuild/plugin-node-polyfill": "^1.0.4",
    "@rsbuild/plugin-sass": "^1.0.1-rc.0",
    "@rsbuild/plugin-vue2": "^1.0.1",
    "@rsbuild/plugin-vue2-jsx": "^1.0.1-beta.14",
    "@rsdoctor/rspack-plugin": "^0.3.11",
```

### Details

如图所示：![image](https://github.com/user-attachments/assets/57076426-b963-4402-926c-e8bdad77b334)，已经安装了rspack-plugin，应该不需要再提示了。

### Reproduce link

无

### Reproduce Steps

无

chenjiahan:
I have added a fix for this issue., can you try the latest RC version? https://github.com/web-infra-dev/rsbuild/releases/tag/v1.0.1-rc.4

melMass:
Neither rc4 or rc5 solves it for me on windows with pnpm

melMass:
Here is the reproduction: https://github.com/melMass/reproduction/tree/js/rsbuild-rsdoctor

easy1090:
@melMass I wasn't able to reproduce the issue using the reproduction. Could you try deleting the node_modules directory and then run pnpm install again

<img width="843" alt="image" src="https://github.com/user-attachments/assets/74762edd-2b59-4e9e-8453-c81e19728c13">

chenjiahan:
@easy1090 are you using Windows?

easy1090:
@chenjiahan  I didn't notice Windows. Let me try it on Windows.

chenjiahan:
I guess this issue is similar to this one: https://github.com/rspack-contrib/rsbuild-plugin-ejs/pull/2/files

Using `pathToFileURL` should fix it.

PR #3370, opened 2024-09-03
fix: skip resolution if Rsdoctor plugin is registered

chenjiahan:
## Summary

If Rsdoctor plugin has been registered, Rsbuild does not need to resolve the Rsdoctor plugin package.

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/3358

## Checklist

### DIFF
--- a/packages/core/src/plugins/rsdoctor.ts
+++ b/packages/core/src/plugins/rsdoctor.ts
@@ -73,7 +73,6 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
         return;
       }
 
-      // TODO: support for multiple compiler
       for (const config of bundlerConfigs) {
         config.plugins ||= [];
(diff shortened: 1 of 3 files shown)


packages/core/src/plugins/rsdoctor.ts
```diff
@@ -24,56 +24,62 @@ export const pluginRsdoctor = (): RsbuildPlugin => ({
...
         return;
       }
 
-      let isAutoRegister = false;
-
-      const isRsdoctorPlugin = (plugin: MaybeRsdoctorPlugin) =>
-        plugin?.isRsdoctorPlugin === true ||
-        plugin?.constructor?.name === pluginName;
-
+      // TODO: support for multiple compiler
```
chenjiahan, line 76:
```suggestion
```

PR #3434, opened 2024-09-10
feat(plugin-stylus): support `define` option

### LINKED ISSUE #3430: [Feature]: More Stylus options

Trying to migrate from webpack to rsbuild here but stuck:

Webpack's stylus loader has the `define` option:
https://webpack.js.org/loaders/stylus-loader/#stylusoptions

```ts
type stylusOptions =
  | {
      use: Array<string | Function>;
      include: Array<string>;
      import: Array<string>;
      define: Array; // <------------------------------ this one is missing
      includeCSS: false;
      resolveURL: boolean | Object;
      lineNumbers: boolean;
      hoistAtrules: boolean;
      compress: boolean;
    }
  | (loaderContext: LoaderContext) => Array<string>;
  ```

In my app, this enables to pass over some variables from the code to the styles which is extremely useful, such as media sizes.

If this could be added, that would be great. Thanks.

Presently, rsbuild has these types under @rsbuild/plugin-stylus/dist-types/index.d.ts:
```ts
type StylusOptions = {
    use?: string[];
    include?: string[];
    import?: string;
    resolveURL?: boolean;
    lineNumbers?: boolean;
    hoistAtrules?: boolean;
};
```

How about adding a new option named and typed the same way, `define`, allowing arrays? And in my code it would look like that:

```ts
define: [
  ["video-width", getVideoWidth(), true],
  ["video-height", getVideoHeight(), true],
],
```

shulaoda:
## Related Links

closes #3430 

## Checklist

### DIFF
--- a/packages/plugin-stylus/src/index.ts
+++ b/packages/plugin-stylus/src/index.ts
@@ -6,6 +6,7 @@ export const PLUGIN_STYLUS_NAME = 'rsbuild:stylus';
 
 type StylusOptions = {
   use?: string[];
+  define: Array<any>;
   include?: string[];
   import?: string;
   resolveURL?: boolean;
--- a/packages/plugin-stylus/src/index.ts
+++ b/packages/plugin-stylus/src/index.ts
@@ -6,7 +6,7 @@ export const PLUGIN_STYLUS_NAME = 'rsbuild:stylus';
 
 type StylusOptions = {
   use?: string[];
-  define: Array<any>;
+  define?: [string, any, boolean?];
   include?: string[];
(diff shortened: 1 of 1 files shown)


shulaoda:
Let me see if I can refine the any type to a more specific type.

chenjiahan (approved):
Nice 👍

binarykitchen:
Nice, thanks. Maybe update documentation as well next time?

shulaoda:
> Nice, thanks. Maybe update documentation as well next time?

Let me do it.

Issue #3418, opened 2024-09-09
[Bug]: `tools.swc.jsc.transform.react` config overrides takes no effect


SoonIter:
### Version

```sh
System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 67.98 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 128.0.6613.120
    Chrome Canary: 130.0.6702.0
    Safari: 17.6
  npmPackages:
    @rsbuild/core: 1.0.1-rc.5 => 1.0.1-rc.5 
    @rsbuild/plugin-react: 1.0.1-rc.5 => 1.0.1-rc.5
```

### Details

1. `rsbuild.config.ts`

```ts
export default defineConfig({
  plugins: [pluginReact()],
  tools: {
    swc: {
      jsc: {
        transform: {
          react: {
            runtime: 'classic',
          },
        },
      },
    },
  },
});
```

2. `DEBUG=rsbuild rsbuild build`

<img width="689" alt="image" src="https://github.com/user-attachments/assets/ea61ec9d-3c32-403d-aba1-f0204b82a779">

### Reproduce link

https://github.com/SoonIter/rsbuild-tools-swc-issue-repro

### Reproduce Steps

https://github.com/web-infra-dev/rsbuild/blob/14c5a9e3bedba919244818792c238a51f2acefe5/packages/core/src/plugins/swc.ts#L158

1. pnpm install
2. DEBUG=rsbuild npx rsbuild build

SoonIter:
overrided by `@rsbuild/plugin-react`

fixed by

```typescript
export default defineConfig({
  plugins: [pluginReact({
    swcReactOptions: {
      runtime: 'classic'
    }
  })],
  tools: {
    // swc(config) {
    //   (((config.jsc ??= {}).transform ??= {}).react ??= {}).runtime =
    //     'classic';
    // },
  },
});
```

SoonIter:
`tools.swc` should execute in the final stage just like `tools.bundlerChain` and `tools.rspack`

so pluginSwc should execute after pluginReact?

modernjs has no entry for modifying the options of pluginReact

@chenjiahan cc

chenjiahan:
This is an order issue, I think we can try to let `pluginReact` to use `tools.swc` instead of using `modifySwcLoaderOptions`

SoonIter:
[...]

PR #3449, opened 2024-09-11
fix(plugin-react): allow configuring `tools.swc` to override react runtime

chenjiahan:
## Summary

Allow configuring `tools.swc` to override react runtime.

## Related Links

resolve https://github.com/web-infra-dev/rsbuild/issues/3418

## Checklist

### DIFF
--- a/packages/core/src/plugins/basic.ts
+++ b/packages/core/src/plugins/basic.ts
@@ -17,7 +17,7 @@ export const pluginBasic = (): RsbuildPlugin => ({
 
   setup(api) {
     api.modifyBundlerChain(
-      (chain, { env, isProd, target, bundler, environment, CHAIN_ID }) => {
+      (chain, { env, isDev, target, bundler, environment, CHAIN_ID }) => {
         const { config } = environment;
 
         chain.name(environment.name);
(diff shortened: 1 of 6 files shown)


packages/core/src/plugins/basic.ts
```diff
@@ -46,7 +46,7 @@ export const pluginBasic = (): RsbuildPlugin => ({
           },
         });
 
-        const usingHMR = !isProd && config.dev.hmr && target === 'web';
+        const usingHMR = isDev && config.dev.hmr && target === 'web';
```
SoonIter, line 49: Is it okay to do this change?
 I see the snapshots of webpack and svgr are not as same as before

SoonIter (approved):

packages/core/src/plugins/basic.ts
SoonIter: These is also one place of `usingHMR` https://github.com/web-infra-dev/rsbuild/blob/a9995acfabc21135143f87e97a4ae66403e7f7cf/packages/plugin-solid/src/index.ts#L37

chenjiahan: Yes, it is expected, we should only enable HMR when `mode` is `development`.

PR #3462, opened 2024-09-12
feat(core): warn leading slash for output.distPath.html and default to `./`

wjw99830:
## Summary
Currently the default value of `output.distPath.html` is `/`. It's relative to `output.distPath.root`.
But other props(like `js`, `css`...) cannot be `/`, because they're relative to the root of file system.

## Related Links
https://rsbuild.dev/config/output/dist-path

## Checklist


chenjiahan (commented):
I think `output.distPath.html` defaults to `/` is a mistake, it should default to `./`.

By design, `output.distPath.*` can only be relative (except for `output.distPath.root`). If we allow all properties to start with `/`,  I would worry that it might confuse some users.

wjw99830:
> I think `output.distPath.html` defaults to `/` is a mistake, it should default to `./`.
Yeah, i think so too.
If we modify the default value to `./`, it will be a breaking change for user.

chenjiahan:
Yes, we don't want to introduce breaking changes, we can maintain compatibility with the previous behavior and print a warning if the user passes an absolute path to `distPath.html`

packages/core/src/constants.ts
```diff
@@ -3,7 +3,7 @@ import { join } from 'node:path';
 // Paths
 // loaders will be emitted to the same folder of the main bundle
 export const ROOT_DIST_DIR = 'dist';
-export const HTML_DIST_DIR = '/';
+export const HTML_DIST_DIR = '';
```
chenjiahan, line 6: can we use `./` as the default value? It looks more explicit

chenjiahan (approved):
Thanks!

PR #3517, opened 2024-09-20
docs: update `source.exclude` usage

9aoy:
## Summary

update `source.exclude` usage. `source.exclude` is only used to specify JavaScript/TypeScript files that do not need to be compiled. This does not mean that these files will not be bundled into the output.
## Checklist

### DIFF
--- a/website/docs/en/config/source/exclude.mdx
+++ b/website/docs/en/config/source/exclude.mdx
@@ -21,7 +21,7 @@ export default {
 
 ### Not compiled vs Not bundled
 
-`source.exclude` is used to specify JavaScript/TypeScript files that do not need to be compiled. This means that these files will not be translated by `swc` or `babel`, but they will still be bundled into the product (if referenced).
+`source.exclude` is used to specify JavaScript/TypeScript files that do not need to be compiled. This means that these files will not be translated by `swc` or `babel`, but they will still be bundled into the outputs (if referenced).
 
 If you want certain files to be ignored and not bundled into the outputs, you can use Rspack's [IgnorePlugin](https://rspack.dev/plugins/webpack/ignore-plugin).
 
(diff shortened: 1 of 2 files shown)


website/docs/zh/config/source/exclude.mdx
```diff
@@ -18,3 +18,25 @@ export default {
 ```
 
 > 参考 [source.include](/config/source/include) 来了解更多。
+
+### 不编译 vs 不打包
+
+`source.exclude` 用于指定不需要编译的 JavaScript/TypeScript 文件。这意味着这些文件不会经过 `swc` / `babel` 转译，但这些文件仍然会被打包到产物中（如果被引用）。
```
chenjiahan, line 24:
```suggestion
`source.exclude` 用于指定不需要编译的 JavaScript/TypeScript 文件。这意味着这些文件不会经过 SWC / Babel 转译，但这些文件仍然会被打包到产物中（如果被引用）。
```

website/docs/en/config/source/exclude.mdx
```diff
@@ -18,3 +18,25 @@ export default {
 ```
 
 > Refer to [source.include](/config/source/include) to learn more.
+
+### Not compiled vs Not bundled
+
+`source.exclude` is used to specify JavaScript/TypeScript files that do not need to be compiled. This means that these files will not be translated by `swc` or `babel`, but they will still be bundled into the outputs (if referenced).
```
chenjiahan, line 24:
```suggestion
`source.exclude` is used to specify JavaScript/TypeScript files that do not need to be compiled. This means that these files will not be translated by SWC or Babel, but they will still be bundled into the outputs (if referenced).
```

chenjiahan (approved):

PR #3550, opened 2024-09-25
fix(plugin-webpack-swc): swcMinimizerPlugin sourceMap config

easy1090:
## Summary
fix(plugin-webpack-swc): the rsbuildConfig's sourcemap should transmit to the swcMinimizerPlugin
## Checklist

### DIFF
--- a/packages/compat/plugin-webpack-swc/src/minimizer.ts
+++ b/packages/compat/plugin-webpack-swc/src/minimizer.ts
@@ -35,15 +35,19 @@ export class SwcMinimizerPlugin {
 
   private name = 'swc-minimizer-plugin';
 
+  private rsbuildSourceMapConfig: NormalizedEnvironmentConfig['output']['sourceMap'];
+
   constructor(options: {
     jsMinify?: boolean | JsMinifyOptions;
     cssMinify?: boolean | CssMinifyOptions;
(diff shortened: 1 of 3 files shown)


packages/compat/plugin-webpack-swc/src/minimizer.ts
```diff
@@ -9,6 +9,7 @@ import color from 'picocolors';
 import { minify, minifyCss } from './binding';
 import { JS_REGEX } from './constants';
 import type { CssMinifyOptions, JsMinifyOptions, Output } from './types';
+import { isUndefined } from 'lodash';
```
chenjiahan, line 12: We want to avoid using `lodash`, how about `foo === undefined`?

packages/compat/plugin-webpack-swc/src/minimizer.ts
```diff
@@ -35,15 +36,19 @@ export class SwcMinimizerPlugin {
 
   private name = 'swc-minimizer-plugin';
 
+  private rsbuildConfigSourceMap: NormalizedEnvironmentConfig['output']['sourceMap'];
```
chenjiahan, line 39: rsbuildConfigSourceMap -> rsbuildSourceMapConfig

chenjiahan (approved):

PR #3674, opened 2024-10-10
fix(core): ensure that  keep url original

Cheese-Yu:
## Summary
When customize `html.publicPath` with url and using `performance.prefetch` changed the parameters what I set.

Eg:
`publicPath: 'https://{{CDN_HOST}}/{{CDN_PATH_PREFIX}}'`, output is  `<link href="https://{{cdn_host}}/%7B%7BCDN_PATH_PREFIX%7D%7D/static/js/async/318.1e5b1319.js" rel="prefetch">`
Because `urlJoin`  use `new URL` changed `publicPath`.
## Checklist

- [ x ] Tests updated (or not required).

### DIFF
--- a/packages/core/tests/helpers.test.ts
+++ b/packages/core/tests/helpers.test.ts
@@ -121,6 +121,7 @@ it('normalizeUrl', () => {
 
 describe('ensureAssetPrefix', () => {
   const ASSET_PREFIX = 'https://www.example.com/static/';
+  const CAPITAL_ASSET_PREFIX = 'https://www.{{CDN}}.com/{{CDN_PATH}}/';
 
   it('should handle relative url', () => {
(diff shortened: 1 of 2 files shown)


packages/core/tests/helpers.test.ts
```diff
@@ -157,6 +158,15 @@ describe('ensureAssetPrefix', () => {
     expect(ensureAssetPrefix('//foo.com/bar.js', '/')).toBe('//foo.com/bar.js');
     expect(ensureAssetPrefix('/bar.js', '//foo.com')).toBe('//foo.com/bar.js');
   });
+
+  it('should keep url original', () => {
```
chenjiahan, line 162:
```suggestion
  it('should keep the original URL', () => {
```

chenjiahan (approved):
LGTM, thank you!

PR #3746, opened 2024-10-17
chore(CI): fix eco-ci ref & release checkout branch

9aoy:
## Summary

- fix eco-ci ref: should use eco-ci  main brach instead of rsbuild target branch.
- fix release checkout branch: should use selected branch

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,13 +42,14 @@ jobs:
           repo: "rsbuild-ecosystem-ci"
           workflow_file_name: "ecosystem-ci-selected.yml"
           github_token: ${{ secrets.REPO_SCOPED_TOKEN }}
-          ref: ${{ github.event.inputs.branch }}
+          ref: "main"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -42,13 +42,14 @@ jobs:
           repo: "rsbuild-ecosystem-ci"
           workflow_file_name: "ecosystem-ci-selected.yml"
           github_token: ${{ secrets.REPO_SCOPED_TOKEN }}
-          ref: ${{ github.event.inputs.branch }}
+          ref: "main"
```
chenjiahan, line 45: Why use main instead of the input branch here

9aoy: It's the rsbuild-ecosystem-ci's reference of the workflow run, not rsbuild ref. the input branch should be passed in `client_payload`.
https://github.com/convictional/trigger-workflow-and-wait?tab=readme-ov-file#arguments

chenjiahan: get

chenjiahan (approved):

Issue #3751, opened 2024-10-17
[Bug]: Module Federation breaks publicPath / assetsPrefix


danantal:
### Version

  System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 AMD Ryzen 9 3950X 16-Core Processor
    Memory: 28.51 GB / 63.92 GB
  Browsers:
    Edge: Chromium (128.0.2739.42)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @rsbuild/core: 1.0.15 => 1.0.15 
    @rsbuild/plugin-react: 1.0.5 => 1.0.5 

### Details

Adding an empty module federation configuration to a basic RSBuild app breaks the publicPath / assetsPrefix by removing the leading slash.

```
moduleFederation: {
  options: {
    name: 'app',
  }
}
```

With moduleFederation:
![Image](https://github.com/user-attachments/assets/174627db-4699-46d2-bd3c-37830b8eaa74)

Without moduleFederation:
![Image](https://github.com/user-attachments/assets/14acef66-0da7-4320-8a76-5f032630386f)

### Reproduce link

https://github.com/danantal/rsbuild-demo

### Reproduce Steps

1. Run pnpm i
2. Run pnpm dev
3. Navigate to localhost:3000/foo/

Expected: page loads correctly
Actual: blank page due to assets referenced incorrectly

If you comment out the `moduleFederation` configuration in `rsbuild.config.ts` then it works as expected.

danantal:
It seems that this is a problem that occurs due to the defaults that rsbuild applies when configuring module federation. According to docs, when using module federation it does set output.PublicPath to 'auto'.

However, even with an explicit `server.base`, and `dev.assetsPrefix` / `output.assetsPrefix`, the underlying configuration passed to `rspack.output.publicPath` is still 'auto'...

Without module federation:
![Image](https://github.com/user-attachments/assets/9777c517-7d6c-40df-8415-7ad88059d834)

With module federation:
![Image](https://github.com/user-attachments/assets/47474d80-d40a-4801-a28c-1456dd6f2379)

chenjiahan:
As mentioned in the documentation, Rsbuild set the default value of Rspack's [output.publicPath](https://rspack.dev/config/output#outputpublicpath) configuration to 'auto' when using `moduleFederation.options`, see https://rsbuild.dev/config/module-federation/options#introduction

[...]

PR #3757, opened 2024-10-18
fix: configure the default assetPrefix for MF apps correctly

chenjiahan:
## Summary

The default asset prefix has been changed when using `moduleFederation.options`. These changes ensure that the assets of remote modules can be requested by consumer apps with the correct URL, and that the SPA URL fallback works as expected.

- Only change the default asset prefix for MF provider apps.
- Only change the default asset prefix in development mode (via `dev.assetPrefix`).
- The default asset prefix has been changed from `auto` to `true` (equivalent to `localhost:<port>`)
- Always prefer to use the `dev.assetPrefix` value set by user.

@nyqykk please take a look, I suppose these changes should be ported to [@module-federation/rsbuild-plugin](https://github.com/module-federation/core/blob/main/packages/rsbuild-plugin/package.json)

## Related Links

resolve: https://github.com/web-infra-dev/rsbuild/issues/3751

## Checklist

### DIFF
--- a/packages/core/src/plugins/moduleFederation.ts
+++ b/packages/core/src/plugins/moduleFederation.ts
@@ -99,16 +99,26 @@ export function pluginModuleFederation(): RsbuildPlugin {
           return;
         }
 
-        // If this is a provider app, Rsbuild should send the ws request to the provider's dev server.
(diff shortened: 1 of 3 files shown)


nyqykk (approved):

nyqykk:
ok i will port this to universe repo

Issue #3786, opened 2024-10-21
[Bug]: Build with `chunkSplit.override` didn't work if key is module name.


Cheese-Yu:
### Version

  System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 102.58 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Browsers:
    Chrome: 129.0.6668.101
    Edge: 130.0.2849.46
    Safari: 15.6.1
  npmPackages:
    @rsbuild/core: ^1.0.1 => 1.0.16 
    @rsbuild/plugin-vue: ^1.0.1 => 1.0.2 

### Details

1. 'vue' rule is not ok, but 'vendors' rule is ok.
```
    chunkSplit: {
      override: {
        chunks: 'all',
        minSize: 30000,
        cacheGroups: {
          vue: {
            name: 'vue-common',
            test: /[\\/]node_modules[\\/](vue|vue-router|vue-i18n|pinia)[\\/]/,
            priority: 100,
          },
          vendors: {
            name: 'my-chunk-vendors',
            test: /[\\/]node_modules[\\/]/,
            priority: 20,
          },
        },
      },
    },
```
2. Add strategy: 'custom', 'vue' rule is ok, 'vendors' rule is ok too.
```
    chunkSplit: {
      strategy: 'custom'
      override: {
        chunks: 'all',
        minSize: 30000,
        cacheGroups: {
          vue: {
            name: 'vue-common',
            test: /[\\/]node_modules[\\/](vue|vue-router|vue-i18n|pinia)[\\/]/,
            priority: 100,
          },
          vendors: {
            name: 'my-chunk-vendors',
            test: /[\\/]node_modules[\\/]/,
            priority: 20,
          },
        },
      },
    },
```

### Reproduce link

https://github.com/Cheese-Yu/rsbuid-reproduction-mini-repo

### Reproduce Steps

run `npm run build`

chenjiahan:
Hi, what's the expected build result, can you elaborate?

Cheese-Yu:
> Hi, what's the expected build result, can you elaborate?

If use rule:
```
  vue: {
    name: 'vue-common',
    test: /[\\/]node_modules[\\/](vue|vue-router|vue-i18n|pinia)[\\/]/,
    priority: 100,
  }
```
build result should contain a file: `vue-common.[hash].js`

PR #3787, opened 2024-10-21
fix(core): override is noneffective if key is module name

Cheese-Yu:
## Summary
I try to fix this issue, but I'm not sure if there are other side effects.
## Related Links
https://github.com/web-infra-dev/rsbuild/issues/3786

## Checklist

### DIFF
--- a/packages/core/src/plugins/splitChunks.ts
+++ b/packages/core/src/plugins/splitChunks.ts
@@ -275,8 +275,8 @@ export const pluginSplitChunks = (): RsbuildPlugin => ({
           chunkSplit.strategy || 'split-by-experience'
         ]({
           defaultConfig,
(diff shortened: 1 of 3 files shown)


packages/core/src/plugins/splitChunks.ts
```diff
@@ -275,8 +275,8 @@ export const pluginSplitChunks = (): RsbuildPlugin => ({
           chunkSplit.strategy || 'split-by-experience'
         ]({
           defaultConfig,
-          override: override || {},
           forceSplittingGroups,
+          override: override || {},
```
chenjiahan, line 279: Is this an effective change? Changing the order of the object keys does not seem to cause any change.

Issue #3810, opened 2024-10-24
[Bug]: lessLoaderOptions.lessOptions.plugins is not supported.


Sang-Sang33:
### Version

System:
    OS: win10
  Browsers:
    Chorme: 124
  npmPackages:
    "@rsbuild/core": "^1.0.16",
    "@rsbuild/plugin-babel": "^1.0.2",
    "@rsbuild/plugin-vue": "^1.0.2",
    "@rsbuild/plugin-vue-jsx": "^1.0.1",

### Details

rsbuild.config.ts

```typescript
import { TestPlugin } from './test-plugin';
import { resolve } from 'path';
import { defineConfig, rspack } from '@rsbuild/core';
import { pluginLess } from '@rsbuild/plugin-less';
import { pluginVue } from '@rsbuild/plugin-vue';

export default defineConfig({
  plugins: [
    pluginLess({
      lessLoaderOptions: {
        lessOptions: {
          javascriptEnabled: true,
          plugins: [new TestPlugin()]
        },
        implementation: require.resolve('less')
      }
    }),
    pluginVue()
  ],
  html: {
    template: './index.html'
  },
  source: {
    entry: {
      index: './src/main.ts'
    },
  },
});
```
test-plugin.ts
```typescript
export class TestPlugin {
  options?: any;

  constructor(someOptions?: any) {
    this.options = someOptions;
  }

  install(_less: any, manager: any) {
    console.log('manager: ', manager);
    console.log('_less: ', _less);
    console.log(this.options)
   // ....... some code
  }
}
```

Howerver, the plugin's install function has never been invoked when i run `rsbuild build` in the command line.

### Reproduce link

Please see the details.

### Reproduce Steps

Please see the details.

Sang-Sang33:
```typescript
```

PR #3815, opened 2024-10-24
fix(plugin-less): lessLoaderOptions.lessOptions.plugins has lose it's prototype.

Sang-Sang33:
## Summary

fix the `deepmerge` to judge whether is an plain object.

## Related Links

#3810 

## Checklist

### DIFF
--- a/packages/plugin-less/src/index.ts
+++ b/packages/plugin-less/src/index.ts
@@ -5,6 +5,7 @@ import type {
   RsbuildPlugin,
   Rspack,
 } from '@rsbuild/core';
+import { __internalHelper } from '@rsbuild/core';
 import deepmerge from 'deepmerge';
 import { reduceConfigsWithContext } from 'reduce-configs';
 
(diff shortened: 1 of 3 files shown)


packages/plugin-less/src/index.ts
```diff
@@ -75,7 +75,11 @@ const getLessLoaderOptions = (
   ): LessLoaderOptions => {
     const getLessOptions = () => {
       if (defaults.lessOptions && userOptions.lessOptions) {
-        return deepmerge(defaults.lessOptions, userOptions.lessOptions);
+        return deepmerge(defaults.lessOptions, userOptions.lessOptions, {
+          isMergeableObject: (obj) =>
```
chenjiahan, line 79: Can we use the internal `isPlainObject` helper here?

The same as https://github.com/web-infra-dev/rsbuild/blob/650004cee1376f9264673e61a79ff947b964587a/packages/core/src/helpers/index.ts#L72

Sang-Sang33: yeah, of course.

chenjiahan:
I will merge this PR and fix the snapshot, thank you

Issue #3814, opened 2024-10-24
[Feature]: pluginLess cannot be called repeatedly


729993031:
### What problem does this feature solve?

![Image](https://github.com/user-attachments/assets/ce5bb5cf-6cef-4053-84d8-c2dcf14105ec)

## Beground
https://juejin.cn/post/7088140437004746782#heading-9
## Want

Apply different less configurations for different project folders

### What does the proposed API look like?

`  pluginLess({
      lessLoaderOptions: {
        implementation: require('less-old'),
      }
    }),
    pluginLess({
    }),`

chenjiahan:
Rsbuild core can allow the Rsbuild plugin to be registered multiple times, but the `pluginLess` still cannot be called repeatedly, because it uses a fixed Rspack rule id inside.

https://github.com/web-infra-dev/rsbuild/blob/main/packages/plugin-less/src/index.ts#L111-L112

chenjiahan:
Supported in v1.1.0

PR #3818, opened 2024-10-24
fix: allow Rsbuild plugin to be registered times

chenjiahan:
## Summary

Allow Rsbuild plugin to be registered times. This gives plugins more flexibility in designing their APIs, and allows multiple registrations with different options.

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/3814

## Checklist


chenjiahan:
!eco-ci

PR #3860, opened 2024-10-29
fix(plugin/assets-retry): make loadScript safer

SoonIter:
## Summary

😂 loadScript in webpack runtime is for all the script, I'd better to scope it for safer

## Checklist


chenjiahan (approved):
LGTM 👍

PR #3943, opened 2024-11-12
feat: update the assignment method of host in baseUrl

junxiongchu:
## Summary
When local development is allowed and the browser is automatically opened, the host of baseUrl uses the server host configuration by default, instead of directly defaulting to hard-coding localhost
(although it can be specified through open, it is more troublesome for those who use the default open configuration)
## Checklist


chenjiahan:
Thank you for your contribution!

I think this fix is necessary, so I will merge this PR and make it work as expected.

Issue #3781, opened 2024-10-21
[Feature]: Multi-Environment Builds suport config dev.writeToDisk


me9rez:
### What problem does this feature solve?

During the electron application development phase, the main process code needs to be written to disk, but the rendering process does not need to be written to disk. It is hoped that the dev.writeToDisk option can be configured separately according to different environments. Currently, rsbuild v1.0.16 only supports global configuration of this option.

electron应用开发阶段时需要将主进程代码写入到磁盘，但渲染进程可以不写入磁盘，希望可以允许根据不同环境分开配置dev.writeToDisk选项，目前rsbuild v1.0.16版本仅支持全局配置该选项

### What does the proposed API look like?

![Image](https://github.com/user-attachments/assets/cbe52126-b23e-4495-9815-5a3c65a02a3c)

chenjiahan:
@9aoy Is it possible for the Rsbuild dev server to support this?

9aoy:
It needs to be supported in `rsbuild-dev-middleware`, I can take a look.

PR #3946, opened 2024-11-12
feat: allow to configure dev.writeToDisk in environments

9aoy:
## Summary

Allow to configure `dev.writeToDisk` in environments.

```ts
export default defineConfig({
  environments: {
    web1: {
      dev: {
        writeToDisk: true,
      },
    },
    web: {
      dev: {
        writeToDisk: false,
      }
    },
  }
});
```

## Related Links

fix: https://github.com/web-infra-dev/rsbuild/issues/3781

## Checklist


packages/core/src/server/compilerDevMiddleware.ts
```diff
@@ -15,11 +21,54 @@ const require = createRequire(import.meta.url);
...
+
+      return typeof writeToDisk === 'function'
+        ? writeToDisk(filePath)
+        : writeToDisk!;
+    };
+  }
+
+  return newDevConfig;
+};
+
```
chenjiahan, line 71: How about using `Set` to simplify the function:

```suggestion
const formatDevConfig = (
  config: OriginDevConfig,
  environments: Record<string, EnvironmentContext>,
): DevConfig => {
  const writeToDiskValues = Object.values(environments).map(
    (env) => env.config.dev.writeToDisk,
  );
  if (new Set(writeToDiskValues).size === 1) {
    return config;
  }

  return {
    ...config,
    writeToDisk(filePath: string, compilationName?: string) {
      let { writeToDisk } = config;
      if (compilationName && environments[compilationName]) {
        writeToDisk =
          environments[compilationName].config.dev.writeToDisk ?? writeToDisk;
      }
      return typeof writeToDisk === 'function'
        ? writeToDisk(filePath)
        : writeToDisk!;
    },
  };
};
```

chenjiahan (approved):

e2e/cases/server/write-to-disk-environments/index.test.ts
```diff
@@ -0,0 +1,58 @@
+import fs from 'node:fs';
+import { join } from 'node:path';
+import { dev } from '@e2e/helper';
+import { expect, test } from '@playwright/test';
+
+const cwd = __dirname;
+
+test('Multiple environments writeToDisk should work correctly', async ({
```
chenjiahan, line 8:
```suggestion
test('multiple environments writeToDisk should work correctly', async ({
```

PR #3986, opened 2024-11-15
feat: support print error module traces

9aoy:
## Summary
support print error module traces when `stats.moduleTrace` is enabled.

Maybe we can enable moduleTrace by default, because it is enabled by default in [preset: 'errors-only'](https://github.com/web-infra-dev/rspack/blob/f420048ee32b47007b5482b9a6a43721302dcb83/packages/rspack/src/Stats.ts#L186)

```file=rsbuild.config.ts
export default {
  tools: {
    rspack: {
      stats: {
        moduleTrace: true,
      },
    },
};
```
image

## Related Links

resolve: https://github.com/web-infra-dev/rsbuild/issues/3983

## Checklist

### DIFF
--- a/e2e/cases/config/stats-module-trace/index.test.ts
+++ b/e2e/cases/config/stats-module-trace/index.test.ts
@@ -1,0 +1,17 @@
+import { build, proxyConsole } from '@e2e/helper';
+import { expect, test } from '@playwright/test';
+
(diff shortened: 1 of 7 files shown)


chenjiahan (approved):
I agree that `moduleTrace` should be enabled by default, this can be another PR

Issue #3953, opened 2024-11-13
[Bug]: source.include node_modules makes package worker.js bundled


WangLeto:
### Version

```sh
System:
  OS: macOS 15.1
  CPU: (11) arm64 Apple M3 Pro
  Memory: 1.02 GB / 36.00 GB
  Shell: 5.9 - /bin/zsh
Browsers:
  Chrome: 130.0.6723.117
  Edge: 129.0.2792.89
  Safari: 18.1
npmPackages:
  @rsbuild/core: ^1.1.0 => 1.1.0 
  @rsbuild/plugin-react: ^1.0.6 => 1.0.7
```

### Details

[rsbuild-pdfjs-demo-main.zip](https://github.com/user-attachments/files/17725824/rsbuild-pdfjs-demo-main.zip)
 
There is a project zip for reproduction. My goal is to compile the worker.js from the indirect dependency pdfjs-dist, for better browser compatibility, but the `source.input` configuration causes it to be bundled with the main JS asset.

To restore a standard worker.js usage, comment out the `source.input` configuration item.

### Reproduce link

https://github.com/user-attachments/files/17725824/rsbuild-pdfjs-demo-main.zip

### Reproduce Steps

```shell
pnpm i
pnpm dev
```
1. See the network request, worker.js disappeared.
2. Comment out the `source.input` configuration
3. worker.js restored

chenjiahan:
When the `pdf.worker.mjs` is added to the `source.include`, it will matches the JS transform rule and adds the "javascript/auto" module type, so that Rspack will treat it as a JavaScript module rather than a static asset.

GRAMMAC1:
add
```ts
source: {
    include: [{ not: /[\\/]core-js[\\/]|[\\/]pdfjs-dist[\\/]/ }],
 }```
to `rsbuild.config.ts` will work. Based on the docs, Rsbuild will compile the `node_module` code if `source:{inlcude}` is added. Is there any approach to aviod this behaviour? @chenjiahan I tried rspack worker config `worker: ["pdfjs.GlobalWorkerOptions.workerSrc*"]`, but It doesn't work,

chenjiahan:
@GRAMMAC1 The original goal is to "compile the worker.js", so we should not exclude `pdfjs-dist`

PR #4011, opened 2024-11-19
feat: reference script modules as URL assets

chenjiahan:
## Summary

If we use `new URL()` to reference `.js` or `.ts` files, they will be treated as URL assets and will not be processed by Rsbuild's built-in `swc-loader`.

```tsx
// foo.ts will remain the original content and be output to the dist directory
const fooTs = new URL('./foo.ts', import.meta.url).href;

console.log(fooTs); // "/static/foo.[hash].ts"
```

## Related Links

see https://github.com/web-infra-dev/rsbuild/issues/3953

## Checklist

### DIFF
--- a/website/docs/en/guide/basic/static-assets.mdx
+++ b/website/docs/en/guide/basic/static-assets.mdx
@@ -29,6 +29,8 @@ In JS files, you can import static assets with relative paths:
 // Import the logo.png image in the static directory
 import logo from './static/logo.png';
 
+console.log(logo); // "/static/logo.[hash].png"
+
 export default = () => <img src={logo} />;
 ```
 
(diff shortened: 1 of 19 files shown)


website/docs/en/guide/basic/static-assets.mdx
```diff
@@ -37,17 +39,32 @@ Import with [alias](/guide/advanced/alias) is also available:
...
 ```tsx
 const logo = new URL('./static/logo.png', import.meta.url).href;
 
+console.log(logo); // "/static/logo.[hash].png"
+
 export default = () => <img src={logo} />;
 ```
 
+If you use `new URL()` to reference `.js` or `.ts` files, they will be treated as URL assets and will not be processed by Rsbuild's built-in `swc-loader`.
+
```
9aoy, line 60: Shall we need to add a guide on how to treat it as a URL asset and have it processed by the built-in `swc-loader` in Rsbuild?

chenjiahan: I have no idea how to support it, do you have any suggestions?

9aoy: no idea +1 😢

chenjiahan: We can find a solution and include a guide in another PR.

PR #4027, opened 2024-11-20
fix: should close compiler when dev closed

9aoy:
## Summary

Should close compiler when dev closed. According to `webpack-dev-middleware` doc, middleware close only stop watching for file changes.

https://github.com/webpack/webpack-dev-middleware?tab=readme-ov-file#closecallback

It is expected that the relevant memory can be GC when the compiler closed, however currently this part of the memory cannot be fully GC due to unknown reasons.

## Checklist

### DIFF
--- a/packages/core/src/server/compilerDevMiddleware.ts
+++ b/packages/core/src/server/compilerDevMiddleware.ts
@@ -59,9 +59,6 @@ const formatDevConfig = (
     },
   };
 };
-const noop = () => {
-  // noop
-};
 
 function getClientPaths(devConfig: DevConfig) {
   const clientPaths: string[] = [];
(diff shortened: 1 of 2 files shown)


packages/core/src/server/compilerDevMiddleware.ts
```diff
@@ -118,10 +115,24 @@ export class CompilerDevMiddleware {
...
+
+    if (this.middleware) {
+      await new Promise<void>((resolve) => {
+        this.middleware.close(() => {
+          resolve();
+        });
+      });
+    }
+
+    // middleware close only stop watching for file changes.
```
chenjiahan, line 130:
```suggestion
    // `middleware.close()` only stop watching for file changes, compiler should also be closed.
```

chenjiahan (approved):

PR #4074, opened 2024-11-27
feat: always write manifest to disk

9aoy:
## Summary

Always write the manifest to disk so that users can get the manifest from disk, regardless of `dev.writeToDisk` configuration.

## Related Links

https://github.com/rspack-contrib/rspack-examples/pull/178

## Checklist

### DIFF
--- a/packages/core/src/plugins/manifest.ts
+++ b/packages/core/src/plugins/manifest.ts
@@ -143,9 +143,10 @@ export const pluginManifest = (): RsbuildPlugin => ({
   name: 'rsbuild:manifest',
 
   setup(api) {
-    api.modifyBundlerChain(async (chain, { CHAIN_ID, environment }) => {
+    api.modifyBundlerChain(async (chain, { CHAIN_ID, environment, isDev }) => {
       const {
         output: { manifest },
+        dev: { writeToDisk },
       } = environment.config;
 
       if (manifest === false) {
@@ -163,7 +164,7 @@ export const pluginManifest = (): RsbuildPlugin => ({
       chain.plugin(CHAIN_ID.PLUGIN.MANIFEST).use(RspackManifestPlugin, [
         {
           fileName,
-          writeToFileEmit: true,
+          writeToFileEmit: isDev && writeToDisk !== true,
           generate: generateManifest(htmlPaths),
         },
       ]);
--- a/packages/core/src/plugins/manifest.ts
+++ b/packages/core/src/plugins/manifest.ts
@@ -163,6 +163,7 @@ export const pluginManifest = (): RsbuildPlugin => ({
       chain.plugin(CHAIN_ID.PLUGIN.MANIFEST).use(RspackManifestPlugin, [
         {
           fileName,
+          writeToFileEmit: true,
           generate: generateManifest(htmlPaths),
         },
       ]);
--- a/website/docs/en/guide/advanced/ssr.mdx
+++ b/website/docs/en/guide/advanced/ssr.mdx
@@ -152,6 +152,50 @@ If you need to preview the online effect of SSR rendering, you also need to modi
 
 Now, you can run the `npm run dev` command to start the dev server with SSR rendering function, and visit `http://localhost:3000/` to see that the SSR content has been rendered to the HTML page.
 
+## Get Manifest
+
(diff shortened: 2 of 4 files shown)


packages/core/src/plugins/manifest.ts
```diff
@@ -163,6 +163,7 @@ export const pluginManifest = (): RsbuildPlugin => ({
       chain.plugin(CHAIN_ID.PLUGIN.MANIFEST).use(RspackManifestPlugin, [
         {
           fileName,
+          writeToFileEmit: true,
```
chenjiahan, line 166: Will this option conflict with Rspack's built-in emit? Should we enable this only in dev mode

9aoy: There will be no conflict, but it is better to avoid repeated emit.

website/docs/en/guide/advanced/ssr.mdx
```diff
@@ -152,6 +152,50 @@ If you need to preview the online effect of SSR rendering, you also need to modi
...
+  output: {
+    manifest: true,
+  },
+};
+```
+
+```ts title=server.ts
+function renderHtmlPage(): string {
+  const { entries } = JSON.parse(
+    fs.readFileSync('./dist/manifest.json', 'utf-8'),
```
chenjiahan, line 172: can we use `fs.promises`

website/docs/zh/guide/advanced/ssr.mdx
```diff
@@ -152,6 +152,50 @@ startDevServer(process.cwd());
 
 现在，执行 `npm run dev` 命令即可启动带有 SSR 渲染功能的开发服务器，访问 `http://localhost:3000/` 即可看到 SSR 内容已经渲染到了 HTML 页面上。
 
+## 获取资源清单
+
+默认情况下，和当前页面关联的 scripts 和 links 会自动插入到 HTML 模版中，此时通过 [getTransformedHtml](/guide/advanced/environments#environment-api) 即可获取编译后的 HTML 模版内容。
+
+如果你的 HTML 模版是由服务器端提供，此时需要将 web 端的 js 和 css 产物资源注入到 HTML 模版中。使用 [output.manifest](/config/output/manifest) 配置可以轻松获取资源清单。可参考以下示例：
```
chenjiahan, line 159:
```suggestion
如果你的 HTML 模版是由服务器端提供的，此时需要将 web 端的 JS 和 CSS 产物资源注入到 HTML 模版中。使用 [output.manifest](/config/output/manifest) 配置可以轻松获取资源清单。可参考以下示例：
```

website/docs/en/guide/advanced/ssr.mdx
```diff
@@ -152,6 +152,50 @@ If you need to preview the online effect of SSR rendering, you also need to modi
...
+  );
+
+  const { js, css } = entries['index'].initial;
+
+  const scriptTags = js
+    .map((file) => `<script src="${file}" defer></script>`)
```
chenjiahan, line 178:
```suggestion
    .map((url) => `<script src="${url}" defer></script>`)
```

website/docs/en/guide/advanced/ssr.mdx
```diff
@@ -152,6 +152,50 @@ If you need to preview the online effect of SSR rendering, you also need to modi
 
 Now, you can run the `npm run dev` command to start the dev server with SSR rendering function, and visit `http://localhost:3000/` to see that the SSR content has been rendered to the HTML page.
 
+## Get Manifest
+
+By default, scripts and links associated with the current page are automatically inserted into the HTML template. At this time, the compiled HTML template content can be obtained through [getTransformedHtml](/guide/advanced/environments#environment-api).
+
+If your HTML template is provided by the server, you need to inject the web output files into the HTML template. Output manifests can be easily obtained using the [output.manifest](/config/output/manifest) configuration. You can refer to the following example:
```
chenjiahan, line 159:
```suggestion
When you need to dynamically generate HTML on the server side, you'll need to inject the URLs of JavaScript and CSS assets into the HTML. By configuring [output.manifest](/config/output/manifest), you can easily obtain the manifest information of these assets. Here's an example:
```

chenjiahan (approved):

PR #4075, opened 2024-11-27
fix: watchOptions.ignored use array glob instead of regexp

2heal1:
## Summary

Both Webpack and Rspack have the watchOptions.ignore type as `Regexp | string | string[]` , if the internal configuration is Regexp , it's difficult for users to expand this config, so recommend to use array instead.

before 

```ts
const rsbuildConfig = watchOptions.ignored;
watchOptions.ignored = [ transform(rsbuildConfig), userConfig ]
```

after
```ts
const rsbuildConfig = watchOptions.ignored;
watchOptions.ignored = [ ...rsbuildConfig, userConfig ]
```

## Checklist

- [X] Tests updated (or not required).
- [X] Documentation updated (or not required).

### DIFF
--- a/.changeset/sharp-bikes-shout.md
+++ b/.changeset/sharp-bikes-shout.md
@@ -1,0 +1,5 @@
+---
+'@rsbuild/core': patch
+---
+
+fix: watchOptions.ignored use array glob instead of regexp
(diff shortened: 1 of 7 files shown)


chenjiahan (commented):
I actually used RegExp instead of array intentionally for several reasons:

1. I prefer to allow users to "override" the default value rather than "merge with" the default value. Since some users may want to watch `node_modules`, it should be easy for them to override the default behavior.
2. RegExp has better performance. When using glob patterns, Rspack will use [glob-to-regexp](https://github.com/fitzgen/glob-to-regexp) to generate a complex RegExp, which may slow down the build.

2heal1:
gotcha , in fact i just need to modify webpack config , and i find the issue seems to be solved . Okay i will close this pr

PR #4083, opened 2024-11-28
feat(CLI): set default `process.title`

chenjiahan:
## Summary

Set `process.title` to make it easier to identify the process via activity monitor or other tools.

Screenshot 2024-11-28 at 20 45 45

## Related Links

https://github.com/web-infra-dev/rspack/issues/8566
https://nodejs.org/api/process.html#processtitle

## Checklist

### DIFF
--- a/packages/core/src/cli/prepare.ts
+++ b/packages/core/src/cli/prepare.ts
@@ -12,6 +12,9 @@ function initNodeEnv() {
 export function prepareCli(): void {
   initNodeEnv();
 
+  // make it easier to identify the process via activity monitor or other tools
+  process.title = 'rsbuild-node';
+
   // Print a blank line to keep the greet log nice.
   // Some package managers automatically output a blank line, some do not.
   const { npm_execpath } = process.env;


packages/core/src/cli/prepare.ts
```diff
@@ -12,6 +12,9 @@ function initNodeEnv() {
 export function prepareCli(): void {
   initNodeEnv();
 
+  // make it easier to identify the process via activity monitor or other tools
+  process.title = 'rsbuild-node';
```
sibelius, line 16: I got this

```diff
if (process.env.RSPACK_TITLE) {
+	process.title = process.env.RSPACK_TITLE;
+} else {
+	process.title = 'rspack';
+}
```

inside @rspack/cli/bin.js

chenjiahan: 1. Is `process.env.RSPACK_TITLE` necessary? Users can override `process.title` in `rspack.config.ts` without the need for additional env variable.
2. I add `-node` suffix to show that this is a Node.js process.

PR #4113, opened 2024-12-03
docs: add environment API type

9aoy:
## Checklist


website/docs/zh/api/javascript-api/environment-context.mdx
```diff
@@ -1,4 +1,8 @@
-# Environment Context
+# Environment Types
```
chenjiahan, line 1: How about using `Environment API` as title?

chenjiahan (approved):

PR #4129, opened 2024-12-05
fix(plugin-assets-retry): support `output.assetPrefix` with absolute url

SoonIter:
## Summary

let async-retry url calculating aligned with initial-retry

#### Before

only support

```ts
domain: ['https://a.com', 'https://b.com']

// or

domain: ['https://a.com/path-foo', 'https://b.com/path-foo']
```

if you uses below

```ts
domain: ['https://a.com/path-foo', 'https://b.com'],
max: 3

output: {
	assetPrefix: 'https://a.com/path-foo'
}
```

the retry contexts: 

```
0. https://a.com/path-foo/static/js/main.js

1. https://a.com/path-foo/static/js/main.js

2. https://b.com/path-foo/static/js/main.js

3. https://a.com/path-foo/static/js/main.js
```

#### after

```ts
domains: ['https://a.com/path-foo', 'https://b.com'],
max: 3

output: {
	assetPrefix: 'https://a.com/path-foo'
}
```

```
0. https://a.com/path-foo/static/js/main.js

1. https://a.com/path-foo/static/js/main.js

2. https://b.com/static/js/main.js

3. https://a.com/path-foo/static/js/main.js
```

### DIFF
--- a/packages/plugin-assets-retry/src/runtime/initialChunkRetry.ts
+++ b/packages/plugin-assets-retry/src/runtime/initialChunkRetry.ts
@@ -247,7 +247,7 @@ function retry(config: RuntimeRetryOptions, e: Event) {
   const originalQuery =
     target.dataset.rsbuildOriginalQuery ?? getQueryFromUrl(url);
 
-  // this function is same with async chunk retry
+  // this function is the same as async chunk retry
   function getUrlRetryQuery(
     existRetryTimes: number,
     originalQuery: string,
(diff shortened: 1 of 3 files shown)


packages/plugin-assets-retry/src/runtime/initialChunkRetry.ts
```diff
@@ -246,7 +246,12 @@ function retry(config: RuntimeRetryOptions, e: Event) {
   // if the initial request is "/static/js/async/src_Hello_tsx.js?q=1", retry url would be "/static/js/async/src_Hello_tsx.js?q=1&retry=1"
   const originalQuery =
     target.dataset.rsbuildOriginalQuery ?? getQueryFromUrl(url);
-  function getUrlRetryQuery(existRetryTimes: number): string {
+
+  // this function is same with async chunk retry
+  function getUrlRetryQuery(
+    existRetryTimes: number,
+    originalQuery: string,
```
chenjiahan, line 253: Is this `originalQuery` param the same as `const originalQuery` in L247?

SoonIter: yes, these two functions are exactly the same now

chenjiahan: But the `originalQuery` param seems to be unnecessary in this file

chenjiahan (approved):

PR #4199, opened 2024-12-15
docs(website): revise the description

xbzhang2020:
## Summary

Change `stage-optimize` to `optimize`, as `stage-optimize` is not defined in [the list of supported stages](https://rsbuild.dev/plugins/dev/core#process-assets-stages). Therefore, the right term is `optimize`.
## Checklist


chenjiahan (approved):
Thank you!

PR #4268, opened 2024-12-25
fix: should copy publicDir to the environment distDir when multiple environments

9aoy:
## Summary

- Should copy publicDir to the environment distDir when multiple environments, except for nested dist directories.
- The public directory is not copied in the node target environment by default.

## Checklist

### DIFF
--- a/e2e/cases/server/public-dir/publicDir.test.ts
+++ b/e2e/cases/server/public-dir/publicDir.test.ts
@@ -170,6 +170,131 @@ test('should serve publicDir for preview server correctly', async ({
   await rsbuild.close();
 });
 
+test('should copy publicDir to the environment distDir when multiple environments', async () => {
+  await fse.outputFile(join(__dirname, 'public', 'test-temp-file.txt'), 'a');
+
+  const rsbuild = await build({
+    cwd,
+    rsbuildConfig: {
+      environments: {
+        web1: {
+          output: {
+            distPath: {
+              root: 'dist-build-web-1',
+            },
+          },
+        },
+        web2: {
+          output: {
+            distPath: {
+              root: 'dist-build-web-2',
+            },
+          },
+        },
+        node: {
+          output: {
+            target: 'node',
+            distPath: {
+              root: 'dist-build-node',
+            },
+          },
+        },
+      },
+    },
+  });
+  const files = await rsbuild.unwrapOutputJSON();
+  const filenames = Object.keys(files);
+
+  expect(
+    filenames.some((filename) =>
+      filename.includes('dist-build-web-1/test-temp-file.txt'),
+    ),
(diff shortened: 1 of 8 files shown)


chenjiahan (commented):
I think this PR can be split into two PRs. The node target related change is important and we should let users know.

e2e/cases/server/public-dir/publicDir.test.ts
```diff
@@ -170,6 +170,85 @@ test('should serve publicDir for preview server correctly', async ({
   await rsbuild.close();
 });
 
+test('should copy publicDir to the environment distDir when multiple environments', async () => {
```
chenjiahan, line 173: Can we add some cases for dev too? We need to make sure that dev and build have the same result.

website/docs/zh/config/server/public-dir.mdx
```diff
@@ -85,6 +85,13 @@ export default {
 在 dev 构建时，如果你需要拷贝一些静态资源到构建产物目录，可以使用 [output.copy](/config/output/copy) 选项代替。
 :::
 
+#### 多环境
+
+当存在多个环境构建时，会将文件从 public 目录复制到各自的构建产物目录下。当构建产物目录存在嵌套的情况时，仅复制到构建产物父目录下。如：
```
chenjiahan, line 90:
```suggestion
当执行 [多环境构建](/guide/advanced/environments) 时，Rsbuild 会将文件从 public 目录拷贝到各个环境的输出目录下。如果输出目录存在嵌套的情况，则只会拷贝文件到输出目录的根目录下。例如：
```

website/docs/en/config/server/public-dir.mdx
```diff
@@ -84,6 +84,13 @@ Note that setting the value of `copyOnBuild` to false means that when you run `r
 During dev builds, if you need to copy some static assets to the output directory, you can use the [output.copy](/config/output/copy) option instead.
 :::
 
+#### Multiple environments
+
+When there are multiple environment builds, the files will be copied from the public directory to their respective distDir. When distDir are nested, they are only copied to their parent directory. For example:
```
chenjiahan, line 89:
```suggestion
When performing [multi-environment builds](/guide/advanced/environments), Rsbuild copies files from the public directory to the output directory of each environment. If there are nested output directories, files will only be copied to the root of the output directory. For example:
```

e2e/cases/server/public-dir/publicDir.test.ts
9aoy: We didn't copy public dir in dev, only serve public files to server.base(`/`). Therefore dev will not be affected by multiple environments.

chenjiahan (approved):

PR #4291, opened 2024-12-29
fix: do not open browser when open parameter is set to false

esskar:
## Summary

Allows to disable browser opening by setting the open parameter to false.

## Checklist

### DIFF
--- a/packages/core/src/server/open.ts
+++ b/packages/core/src/server/open.ts
@@ -107,11 +107,15 @@ export function resolveUrl(str: string, base: string): string {
 
 const normalizeOpenConfig = (
   config: NormalizedConfig,
-): { targets: string[]; before?: () => Promise<void> | void } => {
+): {
+  targets: string[];
+  before?: () => Promise<void> | void;
+  skip?: boolean;
+} => {
   const { open } = config.server;
 
   if (typeof open === 'boolean') {
-    return { targets: [] };
+    return { targets: [], skip: !open };
   }
   if (typeof open === 'string') {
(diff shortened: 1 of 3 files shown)


packages/core/src/server/open.ts
```diff
@@ -107,11 +107,15 @@ export function resolveUrl(str: string, base: string): string {
...
+): {
+  targets: string[];
+  before?: () => Promise<void> | void;
+  skip?: boolean;
+} => {
   const { open } = config.server;
 
   if (typeof open === 'boolean') {
-    return { targets: [] };
+    return { targets: [], skip: !open };
```
chenjiahan, line 118: This doesn't seem necessary, the server plugin already checks if `open` is false:

https://github.com/web-infra-dev/rsbuild/blob/main/packages/core/src/plugins/server.ts#L15

esskar: Then there must be something else wrong as 

```
import { defineConfig } from '@rsbuild/core';
import { pluginReact } from '@rsbuild/plugin-react';

export default defineConfig({
  plugins: [pluginReact()],
  server: {
    open: false,
  },
});
```
still opens the browser after the server started.

chenjiahan: Are you using the `--open` CLI flag? The CLI flag takes precedence over the configurations,

esskar: oh no, it is part of the package dev script. thanks. sorry for the confusion. will close it.

PR #4313, opened 2025-01-01
fix(asset-retry): avoid using function rest parameters

chenjiahan:
## Summary

Avoid using function rest parameters, use `arguments` instead.

Reasons:

- Reduce the generated ES5 code, `asyncChunkRetry.min.js` from `4,582 bytes` -> `4,389 bytes`
- This allows us to use SWC instead of Babel to transform the code. If we use rest parameters, the SWC helpers will introduce `Array.from`, which is an ES6 API. See https://github.com/web-infra-dev/rsbuild/pull/4312.
- The code looks nicer.

## Related Links

- https://github.com/web-infra-dev/rsbuild/pull/4025

## Checklist


SoonIter (approved):

SoonIter:
LGTM~

PR #4315, opened 2025-01-02
fix(plugin/assets-retry): addQuery and switchDomain should work in async css chunk

### LINKED ISSUE #4306: [Bug]: [plugin-assets-retry] retry asyncChunk css file, addQuery not working
### Version

```sh
System:
    OS: macOS 12.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 106.99 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Browsers:
    Chrome: 131.0.6778.205
    Edge: 118.0.2088.76
    Safari: 17.4.1
  npmPackages:
    @rsbuild/core: 1.1.13 => 1.1.13 
    @rsbuild/plugin-assets-retry: 1.0.7 => 1.0.7 
    @rsbuild/plugin-check-syntax: 1.2.0 => 1.2.0 
    @rsbuild/plugin-eslint: 1.1.0 => 1.1.0 
    @rsbuild/plugin-react: 1.1.0 => 1.1.0 
    @rsbuild/plugin-sass: 1.1.2 => 1.1.2 
    @rsbuild/plugin-type-check: 1.2.0 => 1.2.0
```

### Details

rsbuild use @rsbuild/plugin-assets-retry

config
```
// rsbuild.config.ts
import { pluginAssetsRetry } from '@rsbuild/plugin-assets-retry';

pluginAssetsRetry({
    type: ['script', 'link'],
    inlineScript: true,
    onFail: function (params) {},
    addQuery: function (params) {
        var times = params.times,
            originalQuery = params.originalQuery;
        var key = new Date().getTime();
        var query = times === 3 ? 'retry=' + key + '&isLast=1' : 'retry=' + key;
        return originalQuery ? originalQuery + '&' + query : '?' + query;
    }
})
```

Image

✅ initialChunk "index.14cf4b46.css"，addQuery is working
❌ asyncChunk "moreServices.feb340a2.css"，addQuery not working

### Reproduce link

/

### Reproduce Steps

/

SoonIter:
## Summary

fix

```ts
    // At present, we don't consider the switching domain and addQuery of async CSS chunk
    // 1. Async js chunk will be requested first. It is rare for async CSS chunk to fail alone.
    // 2. the code of loading CSS in webpack runtime is complex and it may be modified by cssExtractPlugin, increase the complexity of this plugin.
```

When I implemented this plugin, rsbuild was just switched from `experiment.css` to `cssExtractPlugin` experimentally, I didn't know much about cssExtract.

after this change, it means that `plugin-assets-retry` can only be used together with `rspack.CssExtractPlugin` and MiniCssExtractPlugin in webpack

😂 by the way, it was also not implemented in "eden"

## Related Links

close #4306

## Checklist

### DIFF
--- a/packages/plugin-assets-retry/src/AsyncChunkRetryPlugin.ts
+++ b/packages/plugin-assets-retry/src/AsyncChunkRetryPlugin.ts
@@ -102,32 +102,31 @@ class AsyncChunkRetryPlugin implements Rspack.RspackPluginInstance {
           ? module.constructorName
           : module.constructor?.name;
 
-        const isPublicPathModule =
-          module.name === 'publicPath' ||
-          constructorName === 'PublicPathRuntimeModule' ||
-          constructorName === 'AutoPublicPathRuntimeModule';
-
-        if (constructorName.includes('CssLoadingRuntimeModule')) {
+        const isCssLoadingRuntimeModule = constructorName === 'CssLoadingRuntimeModule';
+        if (isCssLoadingRuntimeModule) {
           // @ts-ignore
           const source = module.source.source.toString();
           const toCode = source.replace('var fullhref = __webpack_require__.p + href;', 'var fullhref = __webpack_require__.rsbuildAsyncCssRetry ? __webpack_require__.rsbuildAsyncCssRetry(href, chunkId) : (__webpack_require__.p + href);');
-          console.log(toCode, 2222222)
           // @ts-ignore
           module.source.source = Buffer.from(toCode, 'utf-8');
+          return
         }
 
-        if (!isPublicPathModule) {
-          return;
-        }
+        const isPublicPathModule =
+        module.name === 'publicPath' ||
+        constructorName === 'PublicPathRuntimeModule' ||
+        constructorName === 'AutoPublicPathRuntimeModule';
 
-        const runtimeCode = this.getRawRuntimeRetryCode();
+        if (isPublicPathModule) {
+          const runtimeCode = this.getRawRuntimeRetryCode();
 
-        // Rspack currently does not have module.addRuntimeModule on the js side,
-        // so we insert our runtime code after PublicPathRuntimeModule or AutoPublicPathRuntimeModule.
-        if (isRspack) {
-          appendRspackScript(module, runtimeCode);
-        } else {
-          appendWebpackScript(module, runtimeCode);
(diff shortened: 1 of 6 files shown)


packages/plugin-assets-retry/src/AsyncChunkRetryPlugin.ts
```diff
@@ -102,23 +121,38 @@ class AsyncChunkRetryPlugin implements Rspack.RspackPluginInstance {
...
+          constructorName === 'CssLoadingRuntimeModule';
+
+        // https://github.com/web-infra-dev/rspack/blob/734ba4cfbec00ab68ff55bac95e7740fe8228229/crates/rspack_plugin_extract_css/src/runtime/css_load.js#L54
+        if (isCssLoadingRuntimeModule) {
+          modifyRuntimeModule(
+            module,
+            (originSource) =>
+              originSource.replace(
+                'var fullhref = __webpack_require__.p + href;',
+                'var fullhref = __webpack_require__.rsbuildLoadStyleSheet ? __webpack_require__.rsbuildLoadStyleSheet(href, chunkId) : (__webpack_require__.p + href);',
```
chenjiahan, line 134: This seems to be an unsafe replacement, if the source code has changed, this replacement may not work or may break the original code.

SoonIter: there is no better solution at present, it is a part of ancient runtime code from mini-css, 
if changed in rspack, eco-ci would take effect for this

chenjiahan: @JSerFeng cc, can you offer some advice?

SoonIter: This function relies on css-extract, so it meets expectations and will have a higher dependence on css-extract, even if I don't use the replace method, any changes to css-extract runtime will  have an influence to assets-retry

chenjiahan (approved):

packages/plugin-assets-retry/src/AsyncChunkRetryPlugin.ts
chenjiahan: Okay, let's keep track of the impact of runtime code changes through our ecosystem-ci.

PR #4385, opened 2025-01-17
fix(core): `output.assetsInclude` adds a rule but not exposes public api

SoonIter:
## Summary

not regular

image
image

should be `${type}-asset-url`

```
font  font-asset-url
additional  additional-asset-url
```

## Related Links

for Rslib Asset
https://github.com/web-infra-dev/rslib/pull/684

## Checklist

### DIFF
--- a/packages/core/src/configChain.ts
+++ b/packages/core/src/configChain.ts
@@ -99,6 +99,8 @@ export const CHAIN_ID = {
     IMAGE: 'image',
     /** Rule for media */
     MEDIA: 'media',
+    /** Rule for additional asset */
+    ADDITIONAL: 'additional',
     /** Rule for js */
     JS: 'js',
     /** Rule for data uri encoded javascript */
--- a/packages/core/src/configChain.ts
+++ b/packages/core/src/configChain.ts
@@ -99,8 +99,8 @@ export const CHAIN_ID = {
     IMAGE: 'image',
     /** Rule for media */
     MEDIA: 'media',
-    /** Rule for additional asset */
-    ADDITIONAL: 'additional',
+    /** Rule for additional assets */
+    ADDITIONAL_ASSETS: 'additional-assets',
     /** Rule for js */
     JS: 'js',
     /** Rule for data uri encoded javascript */
--- a/packages/core/src/plugins/asset.ts
+++ b/packages/core/src/plugins/asset.ts
@@ -131,7 +131,7 @@ export const pluginAsset = (): RsbuildPlugin => ({
       const { assetsInclude } = config.source;
       if (assetsInclude) {
         const { dataUriLimit } = config.output;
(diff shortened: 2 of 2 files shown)


packages/core/src/configChain.ts
```diff
@@ -99,6 +99,8 @@ export const CHAIN_ID = {
     IMAGE: 'image',
     /** Rule for media */
     MEDIA: 'media',
+    /** Rule for additional asset */
+    ADDITIONAL: 'additional',
```
chenjiahan, line 103: Why not use `additional-assets` here

packages/core/src/plugins/asset.ts
```diff
@@ -131,7 +131,7 @@ export const pluginAsset = (): RsbuildPlugin => ({
       const { assetsInclude } = config.source;
       if (assetsInclude) {
         const { dataUriLimit } = config.output;
-        const rule = chain.module.rule('additional-assets').test(assetsInclude);
+        const rule = chain.module.rule('additional').test(assetsInclude);
```
chenjiahan, line 134: Should use `CHAIN_ID.`

SoonIter: same as above,
font, image does not use CHAIN_ID
`CHAIN_ID` only for public api now

packages/core/src/configChain.ts
SoonIter: is `additional-asets-asset-url` acceptable?

packages/core/src/plugins/asset.ts
chenjiahan: If we are not using `CHAIN_ID` here, why does this PR need to update the `CHIND_ID` constant?

SoonIter: to export publicAPI to users, user can modify these rules by this id

packages/core/src/configChain.ts
SoonIter: after using `additional-assets`
image

packages/core/src/plugins/asset.ts
SoonIter: fixed by changing the code above by using `CHAIN_ID` in every rule

packages/core/src/plugins/asset.ts
```diff
@@ -85,7 +86,7 @@ export const pluginAsset = (): RsbuildPlugin => ({
       };
 
       const createAssetRule = (
-        assetType: 'svg' | 'font' | 'image' | 'media',
+        assetType: 'svg' | 'font' | 'image' | 'media' | 'assets',
```
chenjiahan, line 89: The 'assets' type appear to be unused.

SoonIter: removed

PR #4389, opened 2025-01-17
feat(plugin-sass): add `useOriginalUrlResolver` option to disable `resolve-url-loader`

junhea:
## Summary

Adds new `useOriginalUrlResolver` option to disable `resolve-url-loader`

## Related Links

Relative url usage like [this](https://github.com/junhea/rsbuild-sass-issue-sample/blob/2f5df5922c2161fdff50fb1bbec18b1a5a9360fb/src/assets/partials/_foo.scss#L2) should build

## Checklist


chenjiahan (approved):
Thanks for the contribution, providing an option to opt-out resolve-url-loader might be useful. I want to redesign the naming of this API, I will submit a PR to update it later.

PR #4436, opened 2025-01-26
test: fix overlay-type-errors case timeout

9aoy:
## Summary

Should gotoPage after ts check done. Because ts check is asynced in dev, and  reports issues after Rspack's compilation is done.

## Checklist

### DIFF
--- a/e2e/cases/server/overlay-type-errors/index.test.ts
+++ b/e2e/cases/server/overlay-type-errors/index.test.ts
@@ -1,15 +1,41 @@
 import { dev, proxyConsole } from '@e2e/helper';
 import { expect, test } from '@playwright/test';
+import { pluginTypeCheck } from '@rsbuild/plugin-type-check';
 
 const cwd = __dirname;
 
 test('should display type errors on overlay correctly', async ({ page }) => {
   const { restore } = proxyConsole();
 
+  let tsCheckDoneResolve: () => void;
+  const waitTsCheckDone = new Promise<void>((resolve) => {
+    tsCheckDoneResolve = resolve;
+  });
+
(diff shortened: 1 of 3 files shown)


e2e/cases/server/overlay-type-errors/index.test.ts
```diff
@@ -1,16 +1,42 @@
...
+  const waitTsCheckDone = new Promise<void>((resolve) => {
+    tsCheckDoneResolve = resolve;
+  });
+
   const rsbuild = await dev({
     cwd,
-    page,
+    plugins: [
+      {
+        name: 'remove-pre-ts-check-plugin',
```
chenjiahan, line 19: Why not remove the `rsbuild.config.ts` file from this test case folder

chenjiahan (commented):
I think this change goes against the original intention of this case.

In the original case, we open the page first, then wait for the web socket to send type errors to client and display the type errors.

If we wait for the type errors before opening the page, the web socket logic cannot be verified any more.

e2e/cases/server/overlay-type-errors/index.test.ts
```diff
@@ -8,9 +8,13 @@ test('should display type errors on overlay correctly', async ({ page }) => {
 
   const rsbuild = await dev({
     cwd,
-    page,
+    waitFirstCompileDone: false,
   });
 
+  const url = getHrefByEntryName('index', rsbuild.port);
+
+  await page.goto(url, { waitUntil: 'domcontentloaded' });
```
chenjiahan, line 16: Sorry I'm not sure what this change does, could you provide some description?

Issue #4435, opened 2025-01-25
[Bug]: Upgrade from v1.1.14 to 1.2.0 is breaking the stylus plugin


binarykitchen:
### Version

```sh
System:
    OS: Linux 6.12 EndeavourOS
    CPU: (8) x64 Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
    Memory: 5.98 GB / 15.46 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh

  Browsers:
    Chromium: 132.0.6834.110

  npmPackages:
    @rsbuild/core: 1.2.0 => 1.2.0 
    @rsbuild/plugin-node-polyfill: 1.2.0 => 1.2.0 
    @rsbuild/plugin-pug: 1.0.2 => 1.0.2 
    @rsbuild/plugin-react: 1.1.0 => 1.1.0 
    @rsbuild/plugin-stylus: 1.0.7 => 1.0.7 
    @rsbuild/plugin-svgr: 1.0.6 => 1.0.6
```

### Details

Until v1.1.14 I had no issues with your @rsbuild/plugin-stylus plugin. With the bump to v1.2.0 I can't use stylus at all anymore. It now believes it's an invalid javaScript file :(

See error output below:

```zsh
error   Compile error: 
Failed to compile, check the errors for troubleshooting.
/home/michael-heuberger/code/binarykitchen/videomail.io/src/client/assets/styl/site.styl:1:1
  × Module parse failed:
  ╰─▶   × JavaScript parsing error: Expression expected
         ╭─[2:0]
       1 │ // load constants first
       2 │ @import './../../../common/assets/styl/imports/constants/colors'
         · ─
       3 │ @import './../../../common/assets/styl/imports/constants/dimensions'
       4 │ @import './../../../common/assets/styl/imports/constants/fonts'
         ╰────
      
  help: 
        To enable support for Stylus, use "@rsbuild/plugin-stylus".

 @ ./src/client/index.tsx

error   Failed to build.
error   Rspack build failed!
    at file:///home/michael-heuberger/code/binarykitchen/videomail.io/node_modules/@rsbuild/core/dist/index.js:5534:87
    at finalCallback (/home/michael-heuberger/code/binarykitchen/videomail.io/node_modules/@rsbuild/core/node_modules/@rspack/core/dist/index.js:13712:9)
[...]

PR #4437, opened 2025-01-26
fix(core): failed to register environment plugin

chenjiahan:
## Summary

Fix failed to register environment plugin.

Since https://github.com/web-infra-dev/rsbuild/pull/4283, `rsbuildOptions.environment` will be empty array by default, and we should use `context.specifiedEnvironments` to filter environments.

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/4435

## Checklist

### DIFF
--- a/e2e/cases/stylus/environment/src/b.module.styl
+++ b/e2e/cases/stylus/environment/src/b.module.styl
@@ -1,0 +1,2 @@
+.title-class
+  font-size: 14px;
--- a/e2e/cases/stylus/environment/src/index.js
+++ b/e2e/cases/stylus/environment/src/index.js
@@ -1,0 +1,4 @@
+import './a.styl';
+import style from './b.module.styl';
+
+console.log(style);
--- a/packages/core/src/createRsbuild.ts
+++ b/packages/core/src/createRsbuild.ts
@@ -272,15 +272,22 @@ export async function createRsbuild(
   if (rsbuildConfig.environments) {
     await Promise.all(
       Object.entries(rsbuildConfig.environments).map(async ([name, config]) => {
-        const isEnvironmentEnabled =
-          !rsbuildOptions.environment ||
-          rsbuildOptions.environment.includes(name);
-        if (config.plugins && isEnvironmentEnabled) {
-          const plugins = await getFlattenedPlugins(config.plugins);
-          rsbuild.addPlugins(plugins, {
-            environment: name,
-          });
+        if (!config.plugins) {
+          return;
         }
+
+        // The current environment is not specified, skip it
+        if (
+          context.specifiedEnvironments &&
+          !context.specifiedEnvironments.includes(name)
+        ) {
+          return;
+        }
+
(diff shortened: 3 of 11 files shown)


chenjiahan:
@binarykitchen cc

e2e/cases/stylus/environment/src/b.module.styl
```diff
@@ -0,0 +1,2 @@
+.title-class
+  font-size: 14px;
```
binarykitchen, line 2: Small nitpick, in Stylus no need for `:` between property name and value

```
font-size 14px;
````

File: e2e/cases/stylus/environment/src/index.js [hunk 2]
```diff
@@ -0,0 +1,4 @@
+import './a.styl';
+import style from './b.module.styl';
+
+console.log(style);
```
binarykitchen, line 4: It is a good idea to console print during E2E tests?

packages/core/src/createRsbuild.ts
binarykitchen, line 1: Ah, that's the most important change. Curious why this has been left out before?

e2e/cases/stylus/environment/src/b.module.styl
chenjiahan: Nice catch, do you want to submit a PR to polish our stylus code?

packages/core/src/createRsbuild.ts
```diff
@@ -272,15 +272,22 @@ export async function createRsbuild(
...
-        if (config.plugins && isEnvironmentEnabled) {
-          const plugins = await getFlattenedPlugins(config.plugins);
-          rsbuild.addPlugins(plugins, {
-            environment: name,
-          });
+        if (!config.plugins) {
+          return;
         }
+
+        // The current environment is not specified, skip it
```
binarykitchen, line 279: Grammar, add "If" the current environment is not specified ...

e2e/cases/stylus/environment/src/index.js
chenjiahan: To prevent the bundler from removing unused `style' object.

This is client code, so the `console.log` won't affect the E2E logs.

packages/core/src/createRsbuild.ts
chenjiahan: I missed this place, my bad.

e2e/cases/stylus/environment/src/b.module.styl
binarykitchen: Not critical, we developers are very busy :) Next time ...

packages/core/src/createRsbuild.ts
binarykitchen: Appreciate your honesty :bow:

e2e/cases/stylus/environment/src/b.module.styl
chenjiahan: Haha, never mind, I can do it with cursor

Issue #4397, opened 2025-01-20
[Bug]: `output.sourceMap.css: true` does not take effect


Timeless0911:
### Version

```sh
System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 701.70 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 131.0.6778.265
    Safari: 17.6
  npmPackages:
    @rsbuild/core: ^1.1.8 => 1.1.13 
    @rsbuild/plugin-react: ^1.0.7 => 1.1.0
```

### Details

`output.sourceMap.css: true` does not take effect.

When I configure with `output.sourceMap.js`, it works well.

### Reproduce link

basic react demo

### Reproduce Steps

1. Create a basic React project via `npm create rsbuild@latest`
2. Add configuration like:
```diff
import { defineConfig } from "@rsbuild/core";
import { pluginReact } from "@rsbuild/plugin-react";

export default defineConfig({
+  output: {
+   sourceMap: {
+      css: true,
+    },
+ },
  plugins: [pluginReact()],
});

```

9aoy:
Rsbuild will configure the value of `output.sourceMap.js` to rspack's [devtool](https://rspack.dev/config/devtool) configuration. It seems that `devtool` not only affects js, but as a global configuration it also affects the sourceMap generation of other resources. Maybe Rsbuild needs to modify the implementation of `output.sourceMap.js`.

PR #4456, opened 2025-01-29
fix: allow to only generate source map for CSS files

chenjiahan:
## Summary

When JS source map is disabled, but CSS source map is enabled, we should add the `SourceMapDevToolPlugin` to let Rspack generate CSS source map.

## Related Links

- resolve https://github.com/web-infra-dev/rsbuild/issues/4397
- https://rspack.dev/plugins/webpack/source-map-dev-tool-plugin#host-source-maps-externally

## Checklist

### DIFF
--- a/e2e/cases/output/source-map/index.test.ts
+++ b/e2e/cases/output/source-map/index.test.ts
@@ -186,3 +186,28 @@ test('should generate source map correctly in development build', async ({
 
   await rsbuild.close();
 });
+
+test('should allow to only generate source map for CSS files', async () => {
+  const rsbuild = await build({
+    cwd: __dirname,
+    rsbuildConfig: {
+      output: {
+        sourceMap: {
+          js: false,
+          css: true,
+        },
+      },
+    },
+  });
+
+  const files = await rsbuild.unwrapOutputJSON(false);
+
+  const jsMapFiles = Object.keys(files).filter((files) =>
+    files.endsWith('.js.map'),
+  );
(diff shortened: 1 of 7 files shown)


e2e/cases/output/source-map/index.test.ts
```diff
@@ -186,3 +186,28 @@ test('should generate source map correctly in development build', async ({
...
   await rsbuild.close();
 });
+
+test('should allow to only generate source map for CSS files', async () => {
+  const rsbuild = await build({
+    cwd: __dirname,
+    rsbuildConfig: {
+      output: {
+        sourceMap: {
+          js: false,
```
Timeless0911, line 196: I remember that when `output.sourceMap.js` is set to `eval-*` or `inline-*` as well as `output.sourceMap.css: true`, the css sourcemap would not be generated.

chenjiahan: Maybe CSS source maps are also inlined? If not, PR welcome~

Timeless0911: Yeah, this means the type of css sourcemap is determined by which of Javascript devtool configuration. And `eval` devtool can not take effect in css since css can not be wrapped by `eval`, so it is as expected? And for `inline-*` devtool configuration, the css sourcemap is correctly inlined after I verify.

chenjiahan: We can create an Rspack issue to track this, especially as this can be a problem when using experiments.css.

Issue #4466, opened 2025-01-31
[Bug]: HMR Not Working with Rsbuild in Tauri iOS Project


wxiaoyun:
### Version

```sh
System:
    OS: macOS 15.2
    CPU: (12) arm64 Apple M2 Pro
    Memory: 529.88 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 132.0.6834.111
    Safari: 18.2
  npmPackages:
    @rsbuild/core: ^1.2.3 => 1.2.3
    @rsbuild/plugin-babel: ^1.0.3 => 1.0.3
    @rsbuild/plugin-solid: ^1.0.4 => 1.0.4
```

## Description
Hot Module Replacement (HMR) is not functioning when using Rsbuild as the bundler in a Tauri project.
- Tracking: https://github.com/wxiaoyun/expense-tracker/issues/7

## Expected Behavior
- When making changes to the source code, the application should update automatically without requiring a full reload
- HMR should work similarly to how it functions with Vite bundler

![Image](https://github.com/user-attachments/assets/d3a75fed-514f-4fd7-ac69-03133910107c)

![Image](https://github.com/user-attachments/assets/54503ec6-01ea-48cf-907f-ca37c55ced75)

## Actual Behavior
- Changes to source code do not trigger HMR when using Rsbuild
- Application requires a full reload to see changes
- The same setup works correctly when using Vite as the bundler

![Image](https://github.com/user-attachments/assets/de4e6ba5-def5-48a8-abcf-6df041334159)

### Reproduce link

https://github.com/wxiaoyun/rsbuild-hmr-reproduction

### Reproduce Steps

```bash
pnpm i

# Initialize tauri ios development: https://tauri.app/start/prerequisites/#ios
# Setup rust compile targets
rustup target add aarch64-apple-ios x86_64-apple-ios aarch64-apple-ios-sim
# Check ios dev dependencies
pnpm run tauri ios init

# install xcode and download an iOS simulator
# boot one of the simulator
# example:
xcrun simctl boot 56A7D041-EE9F-40BE-8C34-190CA3924DE9

pnpm run tauri:rsbuild ios dev <simulator_device_name> # for rsbuild
pnpm run tauri:vite ios dev <simulator_device_name> # for vite
# e.g. pnpm run tauir:rsbuild ios dev "iPhone 16 Pro"

# open safari devtools to see console logs
```

[...]

PR #4474, opened 2025-02-02
feat: direct ws connection fallback

wxiaoyun:
## Summary

- Fallback to direct ws connection similar to Vite

## Related Links

Resolves https://github.com/web-infra-dev/rsbuild/issues/4466

## Checklist

### DIFF
--- a/packages/core/src/client/hmr.ts
+++ b/packages/core/src/client/hmr.ts
@@ -2,6 +2,7 @@ import type { NormalizedClientConfig } from '../types';
 
 const compilationId = RSBUILD_COMPILATION_NAME;
 const config: NormalizedClientConfig = RSBUILD_CLIENT_CONFIG;
+const resolvedConfig: NormalizedClientConfig = RSBUILD_RESOLVED_CLIENT_CONFIG;
 
 function formatURL({
   port,
@@ -211,13 +212,18 @@ function onClose() {
   removeListeners();
   connection = null;
   reconnectCount++;
-  setTimeout(connect, 1000 * 1.5 ** reconnectCount);
+  setTimeout(() => connect(resolvedConfig), 1000 * 1.5 ** reconnectCount);
 }
 
 // Establishing a WebSocket connection with the server.
-function connect() {
+function connect(cfg?: NormalizedClientConfig) {
+  if (!cfg) {
+    cfg = resolvedConfig;
+    console.info('[HMR] Using direct websocket fallback');
+  }
+
   const { location } = self;
-  const { host, port, path, protocol } = config;
+  const { host, port, path, protocol } = cfg;
   const socketUrl = formatURL({
     protocol: protocol || (location.protocol === 'https:' ? 'wss' : 'ws'),
     hostname: host || location.hostname,
--- a/packages/core/src/client/hmr.ts
+++ b/packages/core/src/client/hmr.ts
@@ -212,7 +212,14 @@ function onClose() {
   removeListeners();
   connection = null;
   reconnectCount++;
-  setTimeout(() => connect(true), 1000 * 1.5 ** reconnectCount);
+  setTimeout(connect, 1000 * 1.5 ** reconnectCount);
+}
+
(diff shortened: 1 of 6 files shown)


packages/core/src/server/hmrFallback.ts
```diff
@@ -0,0 +1,73 @@
...
+> {
+  const [nodeResult, dnsResult] = await Promise.all([
+    dns.lookup('localhost'),
+    dns.lookup('localhost', { verbatim: true }),
+  ]);
+  const isSame =
+    nodeResult.family === dnsResult.family &&
+    nodeResult.address === dnsResult.address;
+  return isSame ? undefined : nodeResult.address;
+}
```
wxiaoyun, line 61: Code is copied from Vite up to line 61

wxiaoyun:
HMR successful on Tauri iOS
CleanShot 2025-02-02 at 22 14 02@2x

packages/core/src/client/hmr.ts
```diff
@@ -211,13 +212,19 @@ function onClose() {
   removeListeners();
   connection = null;
   reconnectCount++;
-  setTimeout(connect, 1000 * 1.5 ** reconnectCount);
+  setTimeout(() => connect(true), 1000 * 1.5 ** reconnectCount);
```
chenjiahan, line 215: I am worried that changing the default fallback logic here will cause some cases to fail to connect.

In Vite, each connect will use the host of the current page first, and then use the direct host if it fails.

However, this PR forces all re-connects to use only the direct host, but it is obvious that it cannot be connected in some scenarios, such as on Cloud IDE.

wxiaoyun: I have changed to onError instead of onClose so that it will break existing reconnect logic

wxiaoyun: image

chenjiahan: Nice, using `onError` looks good to me 👍

packages/core/src/server/hmrFallback.ts
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const wildcardHosts = new Set([
+  '0.0.0.0',
+  '::',
+  '0000:0000:0000:0000:0000:0000:0000:0000',
+]);
+
+export async function resolveHostname(
+  optionsHost: string | boolean | undefined,
```
chenjiahan, line 16: Rsbuild's `server.host` only supports `string` type, maybe we can remove `boolean` type here.

packages/core/src/server/hmrFallback.ts
```diff
@@ -0,0 +1,75 @@
+import { promises as dns } from 'node:dns';
+import { DevConfig, ServerConfig } from 'src/types/config';
```
chenjiahan, line 2: Prefer to use relative path

packages/core/src/server/hmrFallback.ts
chenjiahan: It would be nice if we could do some refactoring based on Rsbuild's needs.

wxiaoyun: Is there any special needs you would like?

packages/core/src/server/hmrFallback.ts
```diff
@@ -0,0 +1,75 @@
...
+    host = undefined; // undefined typically means 0.0.0.0 or :: (listen on all IPs)
+  } else {
+    host = optionsHost;
+  }
+
+  // Set host name to localhost when possible
+  let name = host === undefined || wildcardHosts.has(host) ? 'localhost' : host;
+
+  if (host === 'localhost') {
+    // See #8647 for more details.
```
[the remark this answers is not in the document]
wxiaoyun: Lets remove it

packages/core/src/server/hmrFallback.ts
chenjiahan: the current version is LGTM 👍

Issue #4503, opened 2025-02-08
[Bug]: Modify the dev.watchFiles in plugin is not effective


FaureWu:
### Version

```sh
System:
    OS: macOS 15.1.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 87.22 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 132.0.6834.160
    Edge: 126.0.2592.81
    Safari: 18.1.1
  npmPackages:
    @rsbuild/core: ^1.2.3 => 1.2.4
    @rsbuild/plugin-react: ^1.1.0 => 1.1.0
```

### Details

When I try to add dev.watchFiles listening in the plugin and restart the server when it changes, it doesn't work.

 Below is the code sample. Please see the reproduction steps:

### Reproduce link

https://github.com/FaureWu/rsbuild-watchfiles-bug

### Reproduce Steps

1. pnpm i
2. pnpm run dev
3. Modify needwatch.config.ts
4. The server should normally trigger a restart, but it currently doesn't.

chenjiahan:
Fixed by https://github.com/web-infra-dev/rsbuild/pull/4504

PR #4504, opened 2025-02-08
fix(core): modifying the dev.watchFiles config in plugins is not effective

xbzhang2020:
## Summary

Fix the issue that modifying the `dev.watchFiles` config in plugins is not effective.

In [source code](https://github.com/web-infra-dev/rsbuild/blob/main/packages/core/src/cli/init.ts#L47), the `loadConfig` method can't resolve the config in plugins, which can be resolved by `rsbuild.getNormalizedConfig()`. So changing the occasion for loading `dev.watchFiles` will fix the problem.

## Related Links

related issue #4503 

## Checklist

### DIFF
--- a/packages/core/src/cli/init.ts
+++ b/packages/core/src/cli/init.ts
@@ -44,37 +44,6 @@ export async function init({
       loader: commonOpts.configLoader,
     });
 
-    const command = process.argv[2];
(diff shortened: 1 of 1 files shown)


chenjiahan (approved):
LGTM, thank you!

PR #4509, opened 2025-02-08
docs: fix missing admonition end indicator on inline-assets doc

wralith:
## Summary

Due to the missing `:::` in markdown, _Inline JS files section_ is included _Do you really need to exclude assets from inlining?_ info part of _Force no Inlining_ section.

image

## Related Links

https://rsbuild.dev/guide/optimization/inline-assets#inline-js-files

## Checklist


chenjiahan (approved):
Thank you!

PR #4521, opened 2025-02-10
test(asset): add test for [path] template for Rslib

SoonIter:
## Summary

```ts
export default {
  output: {
    distPath: {
      image: ''
    },
	filename: {
	  image: '[path][name][ext]'
	}
  },
};
```

[path] only supported in generator.options but not implemented for `output.filename` and `output.chunkFilename`

so only for asset, js/jsAsync/css/cssAsync can not use `[path]`

image

## Checklist


packages/core/src/plugins/asset.ts
```diff
@@ -82,6 +82,10 @@ export const pluginAsset = (): RsbuildPlugin => ({
       ) => {
         const distDir = config.output.distPath[assetType];
         const filename = getFilename(config, assetType, isProd);
+        // YES: [path][name] NO: [path]/[name]
+        if (distDir.endsWith('[path]')) {
+          return `${distDir}${filename}`;
```
9aoy, line 87: I'm concerned about whether we need to support `template string` in `output.distPath`. In this case, I think it would be more appropriate to set `output.distPath` to `''` and `output.filename` to `[path][name][ext]`.

SoonIter: > set output.distPath to '' and output.filename to [path][name][ext].

okay, but it is a escape-hatch for users to use Rspack not Rsbuild, lol

9aoy (approved):

PR #4536, opened 2025-02-12
fix(server): remove `.mjs` suffix check when loading ESM module in runner

9aoy:
## Summary

Only rspack outputs require proxy import, so only [experiments.outputModule](https://rspack.dev/config/experiments#experimentsoutputmodule) is required for loading rspack esm outputs in runner, no `.mjs` suffix is ​​required.

## Checklist

### DIFF
--- a/packages/core/src/server/runner/esm.ts
+++ b/packages/core/src/server/runner/esm.ts
@@ -1,4 +1,3 @@
-import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import vm, { type SourceTextModule } from 'node:vm';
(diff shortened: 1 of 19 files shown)


packages/core/src/server/runner/esm.ts
```diff
@@ -6,21 +7,62 @@ import { asModule } from './asModule';
 import { CommonJsRunner } from './cjs';
 import { EsmMode, type RunnerRequirer } from './type';
 
+function findPackageJson(startDir: string) {
+  let dir = path.resolve(startDir);
+  do {
+    const pkgFile = path.join(dir, 'package.json');
+    if (fs.existsSync(pkgFile)) {
```
chenjiahan, line 14: Can we use async fs here

packages/core/src/server/runner/esm.ts
```diff
@@ -6,21 +7,62 @@ import { asModule } from './asModule';
...
+      return pkgFile;
+    }
+    dir = path.join(dir, '..');
+  } while (dir !== path.resolve(dir, '..'));
+  return null;
+}
+
+const ModuleTypeMap = new Map<string, 'ESM' | 'CJS'>();
```
chenjiahan, line 22:
```suggestion
const moduleTypeMap = new Map<string, 'ESM' | 'CJS'>();
```

packages/core/src/server/runner/esm.ts
```diff
@@ -6,21 +7,62 @@ import { asModule } from './asModule';
...
+    return 'CJS';
+  }
+
+  try {
+    const packageJson = findPackageJson(path.dirname(filePath));
+    if (packageJson) {
+      if (ModuleTypeMap.get(packageJson)) {
+        return ModuleTypeMap.get(packageJson);
+      }
+      if (JSON.parse(fs.readFileSync(packageJson, 'utf8')).type === 'module') {
```
chenjiahan, line 39: Use readFile async API

packages/core/src/server/runner/esm.ts
9aoy: No, this method will be executed when static / dynamic import is executed, so a synchronization method is required.

9aoy: Maybe we can have a better way to deal with this, but it will take some extra time 😢

chenjiahan (approved):

PR #4555, opened 2025-02-14
feat: support generate preload link with duplicate

9aoy:
## Summary

By default, if a resource has been added to the current HTML via a script tag, it will not be preloaded additionally.

When set `preload.dedupe` to `false`, Rsbuild will generate preload tags for all eligible resources, even if these resources already exist in the HTML.

```js
export default {
  performance: {
    preload: {
      dedupe: false,
    },
  },
};
```

## Checklist

### DIFF
--- a/website/docs/en/config/performance/preload.mdx
+++ b/website/docs/en/config/performance/preload.mdx
@@ -11,6 +11,7 @@ interface PreloadOption {
   type?: IncludeType;
   include?: Filter;
   exclude?: Filter;
+  duplicate?: boolean;
 }
 ```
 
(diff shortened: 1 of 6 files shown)


website/docs/en/config/performance/preload.mdx
```diff
@@ -111,3 +115,22 @@ export default {
   },
 };
 ```
+
+### preload.duplicate
```
chenjiahan, line 119: How about `dedupe` (true by default)

chenjiahan (approved):

Issue #4563, opened 2025-02-14
[Bug]: Default proxy onError handler is keeping request open in broswer


miindlek:
### Version

```sh
System:
    OS: macOS 15.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 81.33 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 133.0.6943.54
    Safari: 18.3
```

### Details

I encountered a problem with the default onError handler of the server proxy.
At the moment the handler only logs the error. The response is not getting closed (which is normally done by calling res.end("...")). If an error happens, the request will stay open forever in the browser.

Here is the corresponding line:
https://github.com/web-infra-dev/rsbuild/blob/b5e7de0bb14cd11559d87fc15689017b2dd82954/packages/core/src/server/proxy.ts#L33

Since the default onError handler will be used, if no handler is passed, the default error handler of http-proxy-middleware will not be executed.
http-proxy-middleware is handling the error correctly by closing it, see:
https://github.com/chimurai/http-proxy-middleware/blob/1ee26edc1103e96e22b1d38b3de118bc21bfc57e/src/plugins/default/error-response-plugin.ts#L17

If no onError handler is configured, the requests keep staying in the pending state. In the console you can see, that the proxy error was already logged:
<img width="1702" alt="Image" src="https://github.com/user-attachments/assets/5c7c4df5-590c-4766-93a7-eed6eca3df69" />

After setting the onError handler to false, the default http-proxy-middleware error handler is used and the requests correctly resolve to an error:
<img width="1700" alt="Image" src="https://github.com/user-attachments/assets/6ef1106e-f9e3-43b0-a469-98e69bcb7764" />

Setting the onError handler to false to get it to work is not intuitive and also not documented. We should either have a correct default handler, which closes the response, or we should not define a default handler, to let http-proxy-middleware handle it.

### Reproduce link

https://codesandbox.io/p/github/miindlek/rsbuild-codesandbox-example/main?import=true

### Reproduce Steps

1. Open the dev console
2. Click on the Button multiple times.
3. See the network tab. Requests are still hanging.
4. See the proxy already logged the error in the console.

chenjiahan:
@9aoy cc

PR #4577, opened 2025-02-17
fix: should handle proxy error correctly

9aoy:
## Summary

use `http-proxy-middleware` default error handler.

## Related Links

fix: https://github.com/web-infra-dev/rsbuild/issues/4563

## Checklist


chenjiahan (approved):
👍

PR #4578, opened 2025-02-17
fix(core): respect `server.port` when replacing `<port>`

colinaaa:
## Summary

When using Rsbuild without development server(e.g.: `rsbuild.initConfig()` or `rsbuild inspect`), the `publicPath` is not correct since we are using `context.devServer?.port || DEFAULT_PORT`

This patch fix this to use `server.port` when needed.

## Related Links

A fix of #3098.

## Checklist

- [x] Documentation updated (or **not required**).


chenjiahan (approved):
👍

Issue #4491, opened 2025-02-06
[Feature]: Can plugin-assets-retry add a delay to retry? The probability of failure of retrying immediately is a bit high.


303182519:
### What problem does this feature solve?

delay: (times) {
return times* 1000
}

### What does the proposed API look like?

delay: (times) {
return times* 1000
}

PR #4643, opened 2025-02-23
feat(plugin-assets-retry): add delay parameter

hangCode2001:
## Summary
add delay parameter，supports passing in number or function.
image

## Related Links
https://github.com/web-infra-dev/rsbuild/issues/4491

## Checklist


packages/plugin-assets-retry/src/runtime/asyncChunkRetry.ts
```diff
@@ -267,7 +267,7 @@ function ensureChunk(chunkId: string): Promise<unknown> {
 
   callingCounter.count += 1;
 
-  return result.catch((error: Error) => {
+  return result.catch(async (error: Error) => {
```
SoonIter, line 270: `async`

This will cause a large size of compatibility code to be added when SWC does grammatical degradation

hangCode2001: ok, let me think about how to deal with it in another way.

SoonIter:
```suggestion
  return result.catch((error: Error) => {
```

hangCode2001: done

SoonIter (approved):

PR #4702, opened 2025-03-03
feat(core): add html language attribute

lucioreyli:
## Related Links

[https://dequeuniversity.com/rules/axe/4.10/html-has-lang](https://dequeuniversity.com/rules/axe/4.10/html-has-lang)
[https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry)

Required by [PageSpeed Insights](https://pagespeed.web.dev)
CleanShot 2025-03-03 at 14 14 13

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -104,6 +104,7 @@ const getDefaultSourceConfig = (): NormalizedSourceConfig => {
 };
 
 const getDefaultHtmlConfig = (): NormalizedHtmlConfig => ({
+  lang: 'en',
   meta: {
     charset: { charset: 'UTF-8' },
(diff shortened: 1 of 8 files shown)


lucioreyli:
The proposal of this feature is to not create a extra html file, just editing the default entry html file and fixing this accessibility issue

chenjiahan:
Thank you for your contribution!

My concern is that adding a default setting of `lang="en"` might not be friendly for non-English language sites.

Another concern is whether we need to add a configuration for `lang`, as Rsbuild cannot add configuration for all common HTML attributes. Perhaps it is a better way to use [html.template](https://rsbuild.dev/config/html/template) directly to fully control the template.

PR #4745, opened 2025-03-10
feat(plugin-webpack-swc): use require.resolve to check react version to support pnpm catalog

caohuilin:
## Summary

use require.resolve to check react version to support pnpm catalog

## Checklist

### DIFF
--- a/packages/compat/plugin-webpack-swc/src/utils.ts
+++ b/packages/compat/plugin-webpack-swc/src/utils.ts
@@ -103,7 +103,7 @@ const isBeyondReact17 = async (cwd: string) => {
   
     return isVersionBeyond17(reactVersion);
   } catch(error) {
-    console.error('Error resolving React version:', error);
+    console.error('Failed to resolve React version:', error);
     return false;
   }
 };
--- a/packages/compat/plugin-webpack-swc/src/utils.ts
+++ b/packages/compat/plugin-webpack-swc/src/utils.ts
@@ -96,7 +96,16 @@ const isBeyondReact17 = async (cwd: string) => {
     return false;
   }
 
-  return isVersionBeyond17(deps.react);
+  try {
+    const reactPath = require.resolve('react/package.json', { paths: [cwd] });
+
+    const reactVersion = JSON.parse(fs.readFileSync(reactPath, 'utf8')).version;
+  
+    return isVersionBeyond17(reactVersion);
+  } catch(error) {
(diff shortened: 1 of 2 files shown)


packages/compat/plugin-webpack-swc/src/utils.ts
```diff
@@ -96,7 +96,16 @@ const isBeyondReact17 = async (cwd: string) => {
...
 
-  return isVersionBeyond17(deps.react);
+  try {
+    const reactPath = require.resolve('react/package.json', { paths: [cwd] });
+
+    const reactVersion = JSON.parse(fs.readFileSync(reactPath, 'utf8')).version;
+  
+    return isVersionBeyond17(reactVersion);
+  } catch(error) {
+    console.error('Error resolving React version:', error);
```
chenjiahan, line 106:
```suggestion
    console.error('Failedd to resolve React version:', error);
```

packages/compat/plugin-webpack-swc/src/utils.ts
```diff
@@ -96,7 +96,16 @@ const isBeyondReact17 = async (cwd: string) => {
     return false;
   }
 
-  return isVersionBeyond17(deps.react);
+  try {
```
chenjiahan, line 99: Do we still need to read package.json?

image

caohuilin: I feel that it is necessary. The current project must have a React dependency. Otherwise, it is easy to use phantom dependencies.

chenjiahan: get👌

packages/compat/plugin-webpack-swc/src/utils.ts
```diff
@@ -96,7 +96,16 @@ const isBeyondReact17 = async (cwd: string) => {
...
+    const reactPath = require.resolve('react/package.json', { paths: [cwd] });
```
chenjiahan, line 100: `createRequire` should be used in ESM modules:

```js
import { createRequire } from 'node:module';

const require = createRequire(import.meta.url);
```

chenjiahan (dismissed):
👍👍

caohuilin:
@chenjiahan  I fixed the lint error. It needs to be approved again ❤️

chenjiahan (approved):

PR #4775, opened 2025-03-13
feat(commands): allow unknown options for CLI commands

hasnum-stack:
## Checklist


hasnum-stack:
@chenjiahan , hi 大佬我想使用额外的命令行参数传给插件, 能否添加一下该配置?

chenjiahan:
For additional options, I would recommend using env variables such as `FOO=1 rsbuild dev`.

If we enable `allowUnknownOptions` in the Rsbuild CLI, this will cause:

- The user will not be prompted if they use the wrong option
- Rsbuild may cause breaking changes when adding new options

hasnum-stack:
okay, 谢谢指点

PR #4790, opened 2025-03-14
fix(cache): `buildCache.cacheDirectory` config not work

Shucei:
## Summary
Fix: Rspack cache directory config not taking effect due to incorrect config structure

When configuring `performance.buildCache.cacheDirectory` in Rsbuild, the cache directory setting was not being applied correctly for Rspack. This is because Rspack requires the cache directory to be configured under `experiments.cache.storage.directory`, while the current implementation was placing it directly under `experiments.cache.directory`.

Before:
```typescript
chain.experiments({
  cache: {
    type: 'persistent',
    directory: cacheDirectory, // ❌ incorrect placement
    buildDependencies,
    version: cacheVersion
  }
});

## Related Links
Rspack cache configuration documentation: https://rspack.dev/config/cache

<!--- Provide links of related issues or pages -->

## Checklist
Updated the cache directory configuration structure in packages/core/src/plugins/cache.ts

<!--- Check and mark with an "x" -->

### DIFF
--- a/e2e/cases/javascript-api/build-and-load-env/.env
+++ b/e2e/cases/javascript-api/build-and-load-env/.env
@@ -1,0 +1,1 @@
+PUBLIC_FOO=foo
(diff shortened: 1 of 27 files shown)


chenjiahan (dismissed):
Thank you!

chenjiahan (approved):

Issue #4879, opened 2025-03-26
[Feature]: Support BROWSER_ARGS


ianzone:
### What problem does this feature solve?

I'd like to open browser with BROWSER_ARGS, `--inprivate` e.g

### What does the proposed API look like?

respect the BROWSER_ARGS env

PR #4886, opened 2025-03-26
feat: support for BROWSER_ARGS

ianzone:
## Related Links
https://github.com/web-infra-dev/rsbuild/issues/4879

## Checklist


packages/core/src/server/open.ts
```diff
@@ -39,15 +39,14 @@ const getTargetBrowser = async () => {
  * https://github.com/facebook/create-react-app/blob/master/LICENSE
  */
 async function openBrowser(url: string): Promise<boolean> {
+  const targetBrowser = await getTargetBrowser();
```
chenjiahan, line 42: The `getTargetBrowser` method uses `ps cax` and can only be executed under darwin

packages/core/src/server/open.ts
```diff
@@ -72,7 +71,11 @@ async function openBrowser(url: string): Promise<boolean> {
   // (It will always open new tab)
   try {
     const { default: open } = await import('../../compiled/open/index.js');
-    await open(url);
+    const browserArgs = process.env.BROWSER_ARGS?.split(' ');
+    const options = targetBrowser
+      ? { app: { name: targetBrowser, arguments: browserArgs } }
```
chenjiahan, line 76: I'm curious if it's possible to use `process.env.BROWSER_ARGS` alone without specifying `process.env.BROWSER`?

packages/core/src/server/open.ts
ianzone: 函数增加了平台判断

packages/core/src/server/open.ts
ianzone: 根据 open 函数的接口定义，name 是必选项

ianzone:
还有啥要改的吗？

packages/core/src/server/open.ts
```diff
@@ -13,7 +13,9 @@ const supportedChromiumBrowsers = [
   'Google Chrome Dev',
   'Google Chrome Beta',
   'Google Chrome',
+  'chrome',
```
chenjiahan, line 16: A PR should only contain one change. If you need to change other features besides supporting BROWSER_ARGS, please create a separate PR.

ianzone: How could you catch when the process name is just `chrome` or `edge`?

chenjiahan: Sorry I don't understand what you said, could you elaborate on it?

packages/core/src/server/open.ts
```diff
@@ -24,7 +26,12 @@ const getTargetBrowser = async () => {
...
   // If user setting not found or not support, use opening browser first
   if (!targetBrowser || !supportedChromiumBrowsers.includes(targetBrowser)) {
-    const { stdout: ps } = await execAsync('ps cax');
+    const { stdout: ps } =
+      process.platform === 'darwin'
+        ? await execAsync('ps cax')
+        : process.platform === 'win32'
+          ? await execAsync('tasklist')
+          : await execAsync('ps');
     targetBrowser = supportedChromiumBrowsers.find((b) => ps.includes(b));
```
ianzone, line 35: You are finding `supportedChromiumBrowsers` in the process list, but the process name could be just `chrome` or `edge`. How would you match these words if they are not in `supportedChromiumBrowsers`?

chenjiahan: I guess you are not using macOS

image

ianzone: Windows indeed

ianzone:
改了下浏览器名的匹配规则，更适应跨平台的情况

ianzone:
Well, now it should be a quick merge.

chenjiahan (approved):
Thank you

PR #4890, opened 2025-03-27
chore(workflow): should not run eco-ci when only doc changed

9aoy:
## Summary

use `dorny/paths-filter` instead of head commit message check.
## Checklist


chenjiahan (approved):
👍

PR #4898, opened 2025-03-27
fix(server): allow `server.headers` to override `server.cors`

chenjiahan:
## Summary

Allow `server.headers` to override `server.cors`. This ensures that fields explicitly defined in `server.headers` always work as expected.

## Checklist

### DIFF
--- a/packages/core/src/server/devMiddlewares.ts
+++ b/packages/core/src/server/devMiddlewares.ts
@@ -101,7 +101,7 @@ const applyDefaultMiddlewares = async ({
   }
 
   // apply `server.headers` option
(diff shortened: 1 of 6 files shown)


packages/core/src/server/devMiddlewares.ts
```diff
@@ -112,6 +100,18 @@ const applyDefaultMiddlewares = async ({
     );
   }
 
+  // apply `server.headers` option
+  // server.headers` can override `server.cors`
```
chenjiahan, line 104:
```suggestion
  // `server.headers` can override `server.cors`
```

PR #4909, opened 2025-03-28
fix: make CSS preprocessor plugins compat with Rsbuild 1.2

colinaaa:
## Summary

Ensure compatibility of plugin-less, plugin-sass, and plugin-stylus with Rsbuild 1.2.x.

> I believe it might also crash when using Rsbuild v1.2.x with plugin-sass v1.3.0, though I haven’t tested it myself. 😉

## Related Links

See https://github.com/lynx-family/lynx-stack/pull/386.

## Checklist

- [x] **Tests updated** (or not required).
- [ ] Documentation updated (or **not required**).

### DIFF
--- a/packages/plugin-less/src/index.ts
+++ b/packages/plugin-less/src/index.ts
@@ -149,11 +149,6 @@ export const pluginLess = (
         .resolve.preferRelative(true)
         .end();
 
-      const inlineRule = chain.module
-        .rule(findRuleId(chain, CHAIN_ID.RULE.LESS_INLINE))
-        .test(include)
-        .resourceQuery(/inline/);
-
       // Support for importing raw Less files
       chain.module
         .rule(CHAIN_ID.RULE.LESS_RAW)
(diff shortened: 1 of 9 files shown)


packages/plugin-sass/src/index.ts
```diff
@@ -126,11 +126,6 @@ export const pluginSass = (
...
         .end();
 
-      const inlineRule = chain.module
-        .rule(findRuleId(chain, CHAIN_ID.RULE.SASS_INLINE))
-        .test(include)
-        .resourceQuery(/inline/);
-
       // Support for importing raw Sass files
       chain.module
         .rule(CHAIN_ID.RULE.SASS_RAW)
```
colinaaa, line 131: `CHAIN_ID.RULE.SASS_RAW` is `undefined` in v1.2.x. So we will get a rule like this:

```
      /* config.module.rule('undefined') */
      {
        test: /\.s(?:a|c)ss$/,
        resourceQuery: /raw/,
        type: 'asset/source'
      },
```

I'll make a patch in the next PR

```suggestion
        .rule(CHAIN_ID.RULE.SASS_RAW ?? "sass-raw")
```

packages/plugin-less/src/index.ts
```diff
@@ -173,7 +168,15 @@ export const pluginLess = (
         callback: (rule: RspackChain.Rule, type: 'normal' | 'inline') => void,
       ) => {
         callback(rule, 'normal');
-        callback(inlineRule, 'inline');
+
+        // Rsbuild < 1.3.0 does not have RULE.CSS_INLINE.
+        if (chain.module.rules.has(CHAIN_ID.RULE.CSS_INLINE)) {
```
colinaaa, line 173: It might be preferable to use `CHAIN_ID.RULE.CSS_INLINE && chain.module.rules.has(CHAIN_ID.RULE.CSS_INLINE)` because, theoretically, there could be a rule named `undefined`.

chenjiahan: this will be better👍

packages/plugin-less/src/index.ts
```diff
@@ -173,7 +168,15 @@ export const pluginLess = (
...
+          const inlineRule = chain.module
+            .rule(findRuleId(chain, CHAIN_ID.RULE.LESS_INLINE))
```
chenjiahan, line 175: findRuleId should not be put inside the updateRules method. It should be only called once.

chenjiahan (approved):

PR #5048, opened 2025-04-17
Fix `server.open: false` keeps to open page

Profesor08:
## Summary

When `server.open` is set to `false`, then page keeps opening in browser.

Expected: 
- page to not open

Actual behavior:
- page is opening
- condition in the code, sets `open` to be `true` if it matches any falsy value.

## Checklist


chenjiahan:
The current behavior is by design. When using server.open and --open at the same time, --open takes precedence.

See https://rsbuild.dev/config/server/open

Profesor08:
The current behavior is by design

PR #5103, opened 2025-04-25
fix(security): update http-proxy-middleware to 2.0.9

DeveshSapkale:
## Summary
In http-proxy-middleware before 2.0.8 and 3.x before 3.0.4, writeBody can be called twice because "else if" is not used.

## Related Links
https://nvd.nist.gov/vuln/detail/CVE-2025-32996

## Checklist


chenjiahan (approved):
Thank you!

DeveshSapkale:
@chenjiahan When can we expect the release build in the next few days? Is there a rough estimate for when this change will be available on npm?

PR #5167, opened 2025-05-06
perf: enable Rspack incremental build by default

chenjiahan:
## Summary

- Enable Rspack incremental build by default because it is stable enough and can bring HMR performance improvements.
- Removed the `EXPERIMENTAL_RSPACK_INCREMENTAL` environment variable

If you run into any issues, you can switch to the safe and stable incremental, which is consistent with previous behavior, by adding the following configuration:

```js
export default defineConfig({
  tools: {
    rspack: {
      experiments: {
        incremental: 'safe',
      },
    },
  },
});
```

## Related Links

- https://rspack.dev/config/experiments#experimentsincremental

## Checklist

### DIFF
--- a/packages/core/src/plugins/basic.ts
+++ b/packages/core/src/plugins/basic.ts
@@ -100,13 +100,11 @@ export const pluginBasic = (): RsbuildPlugin => ({
         // https://github.com/web-infra-dev/rspack/pull/5486
         process.env.WATCHPACK_WATCHER_LIMIT ||= '20';
 
-        // This is temporary, we will remove it after Rspack incremental is stable
-        if (process.env.EXPERIMENTAL_RSPACK_INCREMENTAL) {
-          chain.experiments({
-            ...chain.get('experiments'),
-            incremental: isDev,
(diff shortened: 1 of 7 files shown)


packages/core/src/plugins/basic.ts
```diff
@@ -100,13 +100,11 @@ export const pluginBasic = (): RsbuildPlugin => ({
...
-        if (process.env.EXPERIMENTAL_RSPACK_INCREMENTAL) {
-          chain.experiments({
-            ...chain.get('experiments'),
-            incremental: isDev,
-          });
-        }
+        // TODO: we can remove it after Rspack incremental is enabled by default
+        chain.experiments({
+          ...chain.get('experiments'),
+          incremental: isDev,
```
ahabhgk, line 106:
```suggestion
          incremental: true,
```
Let's just use true here

chenjiahan: Get 👌

ahabhgk (approved):

PR #5222, opened 2025-05-12
docs: correct default `buildDependencies`

colinaaa:
## Summary

The `.env` and `.env.*` files are included in `buildDependencies` within the `createRsbuild` function. This setup allows these files to be utilized by frameworks built on top of Rsbuild, such as Rspeedy and Rslib.

https://github.com/web-infra-dev/rsbuild/blob/045c4e45dfce3d0d0fc5dc27a8777c5fb1874f4a/packages/core/src/createRsbuild.ts#L140-L150

In contrast, the `rsbuild.config.*` files are added to `buildDependencies` within the `loadConfig` function.

https://github.com/web-infra-dev/rsbuild/blob/045c4e45dfce3d0d0fc5dc27a8777c5fb1874f4a/packages/core/src/config.ts#L449-L452

However, this does not apply to Rspeedy, as it does not utilize the `loadConfig` function from Rsbuild.

It does work in Rslib and Rspress (using `rslib.config.*` and `rspress.config.*`), so I'm uncertain if the current explanation for `rsbuild.config.*` is sufficiently clear.

## Related Links

See https://github.com/lynx-family/lynx-stack/pull/766

## Checklist

- [ ] Tests updated (or **not required**).

### DIFF
--- a/website/docs/en/config/performance/build-cache.mdx
+++ b/website/docs/en/config/performance/build-cache.mdx
@@ -100,11 +100,11 @@ Rsbuild will use the following configuration files as the default build dependen
 
 - `package.json`
 - `tsconfig.json`
-- `rsbuild.config.*`
+- `.env`, `.env.*`
 - `.browserslistrc`
 - `tailwindcss.config.*`
 
-When using Rsbuild CLI, it will also automatically add `.env` and `.env.*` files to the build dependencies.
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
LGTM 👍

When using JavaScript API, the default value depends on the parameters of `createRsbuild()`, I will add more description later.

PR #5289, opened 2025-05-21
docs: correct build profiling output

colinaaa:
## Summary

The default value should be `.rspack-profile-${timestamp}-${pid}`.

## Related Links

https://github.com/web-infra-dev/rsbuild/blob/dcc17432549e356d7b9d4a5f94e1ceec81a7b3ce/packages/core/src/plugins/rspackProfile.ts#L51

Also see https://github.com/lynx-family/lynx-stack/pull/873

## Checklist

- [ ] Tests updated (or **not required**).


9aoy (approved):

9aoy:
👍

PR #5334, opened 2025-05-29
feat: support get manifest via environment API

9aoy:
## Summary

support get manifest via environment API.

```ts
const webManifest = await environments.web.getManifest();

console.log(webManifest.entries);
```

## Checklist

### DIFF
--- a/e2e/cases/output/manifest-environment/index.test.ts
+++ b/e2e/cases/output/manifest-environment/index.test.ts
@@ -126,3 +126,66 @@ test('should allow to access manifest data in environment context after dev buil
 
   await rsbuild.close();
 });
+
+test('should allow to access manifest data in environment API', async ({
+  page,
+}) => {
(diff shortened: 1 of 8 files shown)


9aoy:
> Can we use the plugin hooks to get the manifest?
> 
> See: https://rsbuild.rs/api/javascript-api/environment-api#manifest

If we want to use plugin hooks in devServer to get webManifest, the example is as follows, which is not very convenient.

```ts
let webManifest;
api.onDevCompileDone(({ environments }) => {
  webManifest = environments.web.manifest;
});

// ...
const { } = webManifest;
```

chenjiahan:
Since the `manifest` is already in the environment context, we can expose the environment context instead of adding APIs for each context property.

```js
const { manifest } = await environments.web.context;

console.log(manifest?.entries);
```

chenjiahan (approved):

PR #5353, opened 2025-06-03
feat: allow to get raw JS content with `?raw`

chenjiahan:
## Summary

Add support for importing the raw content of JavaScript, TypeScript, and JSX files through the `?raw` query parameter. This makes the `?raw` query more general and intuitive.

```ts
import rawJs from './script1.js?raw';
import rawTs from './script2.ts?raw';
import rawJsx from './script3.jsx?raw';
import rawTsx from './script4.tsx?raw';

console.log(rawJs); // The raw content of the JS file
console.log(rawTs); // The raw content of the TS file
console.log(rawJsx); // The raw content of the JSX file
console.log(rawTsx); // The raw content of the TSX file
```

## Related Links

https://github.com/web-infra-dev/rsbuild/discussions/2665

## Checklist

### DIFF
--- a/e2e/cases/assets/raw-query/index.test.ts
+++ b/e2e/cases/assets/raw-query/index.test.ts
@@ -2,6 +2,7 @@ import { promises } from 'node:fs';
 import { join } from 'node:path';
 import { build, dev } from '@e2e/helper';
 import { expect, test } from '@playwright/test';
+import { pluginReact } from '@rsbuild/plugin-react';
 import { pluginSvgr } from '@rsbuild/plugin-svgr';
 
 test('should allow to get raw asset content with `?raw` in development mode', async ({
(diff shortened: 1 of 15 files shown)


nanianlisao:
@chenjiahan 

Why not make the distinction directly through oneOf but add an independent rule instead?

I think it should be consistent with the scope of the default rules, for example, also affected by `source.include`. Moreover, using oneOf seems even easier to understand.

Are there any other reasons?

chenjiahan:
Switch to `oneOf` is a breaking change, as it changed the rule structure

chenjiahan:
I'm considering using oneOf in Rsbuild v2. Thank you for your feedback!

nanianlisao:
> I'm considering using oneOf in Rsbuild v2. Thank you for your feedback!

hah, I was just about to ask if I was considering applying it in 2.0

chenjiahan:
Updated in: https://github.com/web-infra-dev/rsbuild/pull/7032

PR #5453, opened 2025-06-20
docs(website): use useI18nUrl to avoid useHookAtTopLevel error

thinkasany:
## Related Links

https://github.com/web-infra-dev/rsbuild/blob/main/website/theme/components/utils.ts#L12C1-L21C2
```
export function useI18nUrl() {
  const lang = useLang();
  const {
    siteData: { lang: defaultLang },
  } = usePageData();
  return useCallback(
    (url: string) => withBase(lang === defaultLang ? url : `/${lang}${url}`),
    [lang, defaultLang],
  );
}
```

The main benefits of this function are:

**Encapsulation of logic:** It encapsulates the logic for generating localized URLs based on the current language and default language, making your components cleaner and more maintainable.

**Hook compliance:**  By returning a memoized function via useCallback, it allows you to use the returned function anywhere in your component (including loops and callbacks) without violating React's hooks rules.

**Performance optimization:**  The returned function is memoized and will only change when lang or defaultLang changes, preventing unnecessary re-renders or recalculations.

**Reusability:**  You can use the returned function to generate multiple URLs within the same component efficiently.

## Checklist

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -64,9 +64,6 @@
       },
       "nursery": {
         "useUniqueElementIds": "off"
-      },
-      "correctness": {
-        "useHookAtTopLevel": "off"
       }
     }
   }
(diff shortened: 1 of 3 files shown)


chenjiahan (approved):
Thank you!

PR #5482, opened 2025-06-25
chore(biome): specify a relative path to the schema inside the package

thinkasany:
## Summary
Maybe it is also a good practice to `specify the relative path of the @biomejs/biome package`. I searched many repositories in github that use this method. The advantage of doing this is that there is no need to manually update the corresponding version.

image
image

## Related Links
https://biomejs.dev/reference/configuration/#schema

## Checklist


chenjiahan (approved):
LGTM 👍

PR #5483, opened 2025-06-25
feat: support output extra configs when `inspectConfig`

9aoy:
## Summary

support output extra configs when `inspectConfig`. For example, some upper-level framework configurations(rstest, etc.).

image

## Checklist


chenjiahan (approved):
LGTM

Issue #5509, opened 2025-06-30
[Bug]: Svelte build just doesn't work


jqueryisamonad:
### Version

```sh
System:
  OS: Linux 6.8 Ubuntu 24.04.1 LTS 24.04.1 LTS (Noble Numbat)
  CPU: (4) x64 AMD EPYC-Milan Processor
  Memory: 2.71 GB / 15.24 GB
  Container: Yes
  Shell: 5.2.21 - /bin/bash
npmPackages:
  @rsbuild/core: ^1.4.0 => 1.4.2 
  @rsbuild/plugin-svelte: ^1.0.10 => 1.0.10
```

### Details

![Image](https://github.com/user-attachments/assets/484d7c06-c4bb-4844-afec-eea3d76d4730)

### Reproduce link

https://codesandbox.io/p/github/rspack-contrib/rsbuild-codesandbox-example/csb-gjgr7m/draft/patient-dawn

### Reproduce Steps

1. Setup project with svelte: `npm create rsbuild@latest`
2. Install some other svelte package: `npm i svelte-sonner`
3. Add `svelte-sonner` import to the `App.svelte`
4. Run `npm run build`

chenjiahan:
This issue can be fixed by setting the build target to es2022+. You can add a `.browserslistrc` file to change the build target.

- .browserslistrc

```
Chrome >= 107
Edge >= 107
Firefox >= 104
Safari >= 16
```

## Ref

- https://stackoverflow.com/a/77992992/6219457
- https://github.com/sveltejs/svelte/issues/11605

PR #5511, opened 2025-06-30
fix(plugin-svelte): correct loader order for transforming *.svelte.js files

chenjiahan:
## Summary

Fix the incorrect loader order for transforming *.svelte.js files.

Run the `svelte-loader` before the `swc-loader`; otherwise, the Svelte compiler may report an error in some cases.

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/5509

## Checklist


chenjiahan:
Closed as it is not the ideal solution. We need to run `swc-loader` to strip types.

PR #5558, opened 2025-07-08
test: speed up UT by external `@rsbuild/core`

9aoy:
## Summary

test: speed up UT by external `@rsbuild/core`.

`@rsbuild/core` has already been built, so there is no need to build it again in test.

## Checklist


chenjiahan:
- before: `pnpm ut  13.69s user 3.06s system 623% cpu 2.685 total`
- after: `pnpm ut  10.67s user 2.36s system 816% cpu 1.595 total`

PR #5587, opened 2025-07-12
fix(types): type error when importing SVG file with React query

GRAMMAC1:
## Summary

Before

3fe95f57ce0bd3382afc64e4d2c26f4f

After

fdecce2522b6d9b56bb37d417fde7f14

type declaration should go in template project since `rsbuild/core` doesn't depend on React

### DIFF
--- a/packages/create-rsbuild/template-react-ts/src/env.d.ts
+++ b/packages/create-rsbuild/template-react-ts/src/env.d.ts
@@ -4,5 +4,4 @@ declare module "*.svg?react" {
   import React from "react";
   const ReactComponent: React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
   export default ReactComponent;
-  export { ReactComponent };
 }
--- a/packages/create-rsbuild/template-react-ts/src/env.d.ts
+++ b/packages/create-rsbuild/template-react-ts/src/env.d.ts
@@ -1,1 +1,8 @@
 /// <reference types="@rsbuild/core/types" />
+
+declare module "*.svg?react" {
+  import React from "react";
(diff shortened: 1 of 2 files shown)


packages/create-rsbuild/template-react-ts/src/env.d.ts
```diff
@@ -1 +1,8 @@
 /// <reference types="@rsbuild/core/types" />
+
+declare module "*.svg?react" {
+  import React from "react";
+  const ReactComponent: React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
+  export default ReactComponent;
+  export { ReactComponent };
```
chenjiahan: I would prefer to remove the named export, since this is not the default behavior of `@rsbuild/plugin-svgr`. When users set `exportType: 'named'`, they can refer to [this guide](https://rsbuild.rs/plugins/list/plugin-svgr#type-declaration) to update the type declaration.

```suggestion
```

packages/create-rsbuild/template-react18-ts/src/env.d.ts
```diff
@@ -1 +1,8 @@
 /// <reference types="@rsbuild/core/types" />
+
+declare module "*.svg?react" {
+  import React from "react";
+  const ReactComponent: React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
+  export default ReactComponent;
+  export { ReactComponent };
```
chenjiahan:
```suggestion
```

chenjiahan (approved):
❤️ Nice!

PR #5604, opened 2025-07-16
chore: remove workspaces filed from root package.json

chenjiahan:
## Summary

Remove workspaces filed from root package.json. This config is used by changeset and we no longer use changeset.

## Checklist


chenjiahan:
check-dependency-version-consistency depends on this.

PR #5608, opened 2025-07-16
chore(plugin-vue): rm unused webpack deps

thinkasany:
## Summary
It looks like webpack is not used, and the product will not change after deleting it.

Or it seems that only one is enough, and there is no need to use both.
## Checklist

### DIFF
--- a/packages/plugin-vue/package.json
+++ b/packages/plugin-vue/package.json
@@ -28,8 +28,7 @@
     "bump": "npx bumpp --no-tag"
   },
   "dependencies": {
-    "vue-loader": "^17.4.2",
-    "webpack": "^5.99.9"
+    "vue-loader": "^17.4.2"
   },
   "devDependencies": {
     "@rsbuild/core": "workspace:*",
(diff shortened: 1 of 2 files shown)


thinkasany:
> Unfortunately, webpack cannot be removed because `vue-loader` still depends on webpack: [vuejs/vue-loader@`main`/src/pluginWebpack5.ts#L12-L14](https://github.com/vuejs/vue-loader/blob/main/src/pluginWebpack5.ts#L12-L14)



https://github.com/nuxt/nuxt/blob/343f59235c899f581a2d6f293d2947be77f9bdcf/packages/webpack/package.json#L67-L68

I took a look at the package `@nuxt/webpack-builder`, and it is true that there are no additional duplicate installs in devDependencies, and it doesn't seem to be duplicated when I look at the lock file.

image

chenjiahan (approved):
It's OK to only remove webpack from devDependencies. 👍

thinkasany:
> It's OK to only remove webpack from devDependencies. 👍



Thanks for posting the source code address, at first I was surprised to see it used in two places at the same time.

I realized there was a problem when copilot commented on it, but in the end I was able to contribute, so that's great!

chenjiahan:
Thank you for your contribution! ❤️

PR #5615, opened 2025-07-17
fix: wait until the last build is finished before calling `stats.toJson`

9aoy:
## Summary

fix rspack panic when call `stats.toJson()`  during rebuild,  rsbuild should wait the last compile done when `loadBundle`. Because, in the next compilation rspack will clean up the old compilation, but `stats.toJson` read from the old compilation, so the panic happens.

## Related Links

https://github.com/web-infra-dev/rspack/issues/11005

## Checklist

### DIFF
--- a/packages/core/src/server/devServer.ts
+++ b/packages/core/src/server/devServer.ts
@@ -161,6 +161,11 @@ export async function createDevServer<
   let waitLastCompileDone: Promise<void> = Promise.resolve();
 
   const resetWaitLastCompileDone = () => {
+    // Resolve the previous promise if it exists
+    if (waitLastCompileDoneResolve) {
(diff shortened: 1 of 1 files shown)


SyMind (approved):

packages/core/src/server/devServer.ts
```diff
@@ -180,6 +189,14 @@ export async function createDevServer<
       );
     }
 
+    compiler?.hooks.run.tap('rsbuild:run', () => {
```
chenjiahan: Do we need to listen `hooks.run` for dev server? I guess `hooks.run` will never be called

9aoy: 👍

chenjiahan (approved):

PR #5696, opened 2025-07-28
fix: remove the deprecated `url.parse` from historyApiFallback

chenjiahan:
## Summary

`url.parse` has been deprecated since Node 24, and we should use `new URL` instead.

Screenshot 2025-07-28 at 21 46 34

## Related Links

- https://nodejs.org/en/blog/release/v24.0.0#deprecations-and-removals
- https://github.com/nodejs/node/pull/55017
- https://github.com/web-infra-dev/rsbuild/pull/5694

## TODO

We still have three dependencies that use `url.parse`:

image

## Checklist

### DIFF
--- a/packages/core/src/types/config.ts
+++ b/packages/core/src/types/config.ts
@@ -1,6 +1,7 @@
 import type { IncomingMessage, ServerResponse } from 'node:http';
 import type { SecureServerSessionOptions } from 'node:http2';
 import type { ServerOptions as HttpsServerOptions } from 'node:https';
+import type { URL } from 'node:url';
 import type {
   Configuration,
   CopyRspackPluginOptions,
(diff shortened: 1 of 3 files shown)


packages/core/src/types/config.ts
```diff
@@ -368,13 +369,12 @@ export type ProxyConfig =
...
 export type HistoryApiFallbackContext = {
   match: RegExpMatchArray;
-  parsedUrl: import('node:url').Url;
+  parsedUrl: URL;
   request: IncomingMessage;
 };
 
 export type HistoryApiFallbackTo =
   | string
-  | RegExp
```
chenjiahan, line 377: Remove the invalid type, see https://github.com/bripkens/connect-history-api-fallback/blob/main/lib/index.js#L93

PR #5753, opened 2025-08-04
feat(core): add `action` option to `rsbuild.initConfigs`

colinaaa:
## Summary

Add `action` option to `rsbuild.initConfigs` to make it easier for testing Rsbuild plugins.

```js
const rsbuild = await createRsbuild();
const [rspackConfig] = await rsbuild.initConfigs({
  action: 'build'
});
```

## Related Links

We are migrating our plugin to use [`apply`](https://rsbuild.rs/plugins/dev/core#conditional-application) in https://github.com/lynx-family/lynx-stack/pull/1310#discussion_r2224826734. And it seems like it's hard to make unit tests since `rsbuild.initConfigs` would **always** apply a plugin due to there is no `context.action`.

## Checklist

### DIFF
--- a/packages/core/src/provider/provider.ts
+++ b/packages/core/src/provider/provider.ts
@@ -50,7 +50,9 @@ export const rspackProvider: RsbuildProvider = async ({
       return build({ context, pluginManager, rsbuildOptions }, options);
     },
 
-    async initConfigs() {
+    async initConfigs(options) {
+      context.action = options?.action;
+
       const { rspackConfigs } = await initConfigs({
         context,
         pluginManager,
--- a/packages/core/src/provider/provider.ts
+++ b/packages/core/src/provider/provider.ts
@@ -51,11 +51,21 @@ export const rspackProvider: RsbuildProvider = async ({
     },
 
     async initConfigs(options) {
+      if (
+        context.action &&
+        options?.action &&
+        context.action !== options.action
+      ) {
+        // Calling initConfigs multiple times with different actions
+        throw new Error(
+          `\
+[rsbuild] initConfigs() can only be called with the same action type.
+  - Expected: ${context.action}
+  - Actual: ${options?.action}`,
+        );
+      }
+
       if (options?.action) {
-        if (context.action !== options.action) {
-          // Calling initConfigs multiple times with different actions
-          delete context.normalizedConfig;
-        }
         context.action = options.action;
       }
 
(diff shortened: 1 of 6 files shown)


packages/core/src/types/rsbuild.ts
```diff
@@ -73,6 +73,16 @@ export type Build = (options?: BuildOptions) => Promise<{
...
 }>;
 
+export type InitConfigsOptions = {
+  /**
+   * The current action type.
+   * - dev: will be set when running `rsbuild dev` or `rsbuild.startDevServer()`
+   * - build: will be set when running `rsbuild build` or `rsbuild.build()`
+   * - preview: will be set when running `rsbuild preview` or `rsbuild.preview()`
+   */
+  action?: RsbuildContext['action'];
```
chenjiahan: How about using `Pick<RsbuildContext, 'action'>` to inherit the JSDoc from `RsbuildContext`

packages/core/src/provider/provider.ts
```diff
@@ -50,7 +50,15 @@ export const rspackProvider: RsbuildProvider = async ({
...
     },
 
-    async initConfigs() {
+    async initConfigs(options) {
+      if (options?.action) {
+        if (context.action !== options.action) {
+          // Calling initConfigs multiple times with different actions
+          delete context.normalizedConfig;
+        }
+        context.action = options.action;
```
chenjiahan, line 69: Is it possible to create a new Rsbuild instance instead of changing the `action` of the same instance?

I'm a bit worried that changing the `action` might lead to unexpected behavior, as Rsbuild currently doesn't support switching between different `action` types on a single instance.

website/docs/en/api/javascript-api/instance.mdx
```diff
@@ -667,6 +677,12 @@ function InitConfigs(): Promise<{
 const rspackConfigs = await rsbuild.initConfigs();
 
 console.log(rspackConfigs);
+
+const buildConfigs = await rsbuild.initConfigs({
```
chenjiahan, line 681: This part can be added to the Chinese documentation too 😄

packages/core/src/provider/provider.ts
colinaaa: Changed to throwing an error when using different `action` on multiple `initConfigs()`.

PR #5790, opened 2025-08-07
perf: update build configs to target ES2022

chenjiahan:
## Summary

Since Rsbuild v1.5 no longer supports Node 16, we can update build configs to target ES2022 to reduce the output code.

### Before

image

### After

image

## Related Links

- https://github.com/web-infra-dev/rsbuild/pull/5786

## Checklist

### DIFF
--- a/packages/core/rslib.config.ts
+++ b/packages/core/rslib.config.ts
@@ -88,7 +88,7 @@ export default defineConfig({
     {
       id: 'esm_index',
       format: 'esm',
-      syntax: 'es2021',
+      syntax: 'es2022',
       plugins: [pluginFixDtsTypes],
       dts: {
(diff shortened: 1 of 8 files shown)


packages/core/rslib.config.ts
```diff
@@ -100,7 +100,7 @@ export default defineConfig({
     {
       id: 'esm_loaders',
       format: 'esm',
-      syntax: 'es2021',
```
fi3ework, line 103: why using `syntax: 'es2022'` instead of `syntax: ['node 18.12.0']` here, any consideration? 🤔

chenjiahan: No, we can switch to node 18.12 here

PR #5796, opened 2025-08-08
tests: adjust rsbuild e2e output tests to fit lazy-compilation

GiveMe-A-Name:
## Checklist

### DIFF
--- a/e2e/scripts/helper.ts
+++ b/e2e/scripts/helper.ts
@@ -2,7 +2,7 @@ import fs from 'node:fs';
 import { platform } from 'node:os';
 import { join } from 'node:path';
 import { stripVTControlCharacters as stripAnsi } from 'node:util';
-import { expect, test } from '@playwright/test';
+import { expect, type Page, test } from '@playwright/test';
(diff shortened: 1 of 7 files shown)


e2e/scripts/helper.ts
```diff
@@ -126,3 +126,22 @@ export const expectPoll = (fn: () => boolean) => {
...
+/**
+ * we enable lazy compilation by default, so we need to wait for the page to load
+ *
+ * This function waits for the HMR (Hot Module Replacement) response to ensure that the page is updated correctly.
+ * It waits for a response that includes "hot-update" in the URL, with a timeout of 5000 milliseconds,
+ * and then waits for an additional 500 milliseconds to ensure the page has time to reflect the changes.
+ * @param page
+ * @param timeout
+ */
+export async function waitForHmr(page: Page, timeout = 100) {
```
chenjiahan, line 139: This is quite confusing, could you explain it?

image

chenjiahan: Waiting for a fixed duration is not a good idea, as the execution time of a test case can vary significantly across different devices or environments.

PR #5812, opened 2025-08-11
chore: integrate with `rsbuild-plugin-arethetypeswrong`

colinaaa:
## Summary

Integrate with [`rsbuild-plugin-arethetypeswrong`](https://www.npmjs.com/package/rsbuild-plugin-arethetypeswrong?activeTab=readme). 

## Related Links

This could avoid future regression on issues like https://github.com/web-infra-dev/rsbuild/pull/3695 or https://github.com/web-infra-dev/rsbuild/pull/3089

## Checklist

### DIFF
--- a/scripts/config/rslib.config.ts
+++ b/scripts/config/rslib.config.ts
@@ -48,6 +48,7 @@ export const dualPackage = defineConfig({
   },
   plugins: [
     pluginAreTheTypesWrong({
+      enable: Boolean(process.env.CI),
       areTheTypesWrongOptions: {
         ignoreRules: [
           // The dual package always provide the ESM types.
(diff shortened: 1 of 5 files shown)


scripts/config/rslib.config.ts
```diff
@@ -45,4 +46,14 @@ export const dualPackage = defineConfig({
       externals: commonExternals,
     },
   },
+  plugins: [
+    pluginAreTheTypesWrong({
+      areTheTypesWrongOptions: {
```
colinaaa: We can only enable this on CI:
```suggestion
    pluginAreTheTypesWrong({
      enable: Boolean(process.env.CI),
      areTheTypesWrongOptions: {
```

chenjiahan: Will this plugin slow down the local build?

colinaaa: A little bit. Since it would use `typescript` for type checking.

chenjiahan: Get. We can only enable this on CI since these kind of errors are relatively rare.

chenjiahan (approved):

chenjiahan:
❤️

PR #5582, opened 2025-07-11
feat: enable lazyCompilation.imports by default

chenjiahan:
## Summary

WIP

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #5824, opened 2025-08-12
feat: enable lazyCompilation.imports by default

GiveMe-A-Name:
## Summary
- Enable lazyCompilation.imports by default.
We need specify lazyCompilation.entries as false , because rspack will enable lazyCompilation.entries if you pass a object as lazyCompilation. more detail see: [Rspack configuration.lazyCompilation](https://rspack.rs/config/lazy-compilation#lazycompilation-1)

- In lazyCompilation, we will do a hmr compile quickly, so we maybe complete the hmr compile before web socket connect done. It means we need do a hmr action when first web socket message send.
## Checklist

### DIFF
--- a/.github/workflows/ecosystem-ci.yml
+++ b/.github/workflows/ecosystem-ci.yml
@@ -16,6 +16,8 @@ permissions:
   contents: write
   # Allow commenting on issues
   issues: write
+  # Allow commenting on PR
+  pull-requests: write
 
 jobs:
   changes:
--- a/.github/workflows/ecosystem-ci.yml
+++ b/.github/workflows/ecosystem-ci.yml
@@ -16,8 +16,6 @@ permissions:
   contents: write
   # Allow commenting on issues
   issues: write
-  # Allow commenting on PR
-  pull-requests: write
 
 jobs:
   changes:
--- a/packages/core/src/client/hmr.ts
+++ b/packages/core/src/client/hmr.ts
@@ -29,7 +29,6 @@ function formatURL(fallback?: boolean) {
 }
 
 // Remember some state related to hot module replacement.
-let isFirstCompilation = true;
 let lastCompilationHash: string | undefined;
 let hasCompileErrors = false;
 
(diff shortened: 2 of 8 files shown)


.github/workflows/ecosystem-ci.yml
```diff
@@ -16,6 +16,8 @@ permissions:
   contents: write
   # Allow commenting on issues
   issues: write
+  # Allow commenting on PR
+  pull-requests: write
```
chenjiahan: Could you submit a separate PR for this change?

packages/core/src/client/hmr.ts
chenjiahan, line 1: Could you explain why you changed the HMR runtime code?

packages/core/src/defaultConfig.ts
```diff
@@ -272,6 +272,14 @@ export const withDefaultConfig = async (
     }
   }
 
+  if (merged.dev?.lazyCompilation === undefined) {
```
chenjiahan, line 275: It would be better if we can update the default config in:

https://github.com/web-infra-dev/rsbuild/pull/5582/files#diff-20fef47b581eb119ad418e27d162f0f73b4e85a2977b374b9ee043985d236cf2R52-R54

.github/workflows/ecosystem-ci.yml
GiveMe-A-Name: Got it

packages/core/src/defaultConfig.ts
GiveMe-A-Name: rsbuild will merge userConfig and defaultConfig. We just enable `lazyCompilation.imports` by default. It means we need specify `lazyCompilation.entries` as `false`. 
But consider this case, if user config lazyCompilation follow bellow:
```js
lazyCompilation: {
   serverUrl: 'xxxx',
   // imports: true,
   // entires: true,
}
```
According to rspack lazyCompilation Configuration rule, if you set it as object rspack will set `imports = true` and `entries = true` by default.

So if we update config in [here](https://github.com/web-infra-dev/rsbuild/pull/5582/files#diff-20fef47b581eb119ad418e27d162f0f73b4e85a2977b374b9ee043985d236cf2R52-R54), user config will become 
```js
lazyCompilation: {
   serverUrl: 'xxxx',
   imports: true,
   entires: false,
}
```
It's will import a break change.

chenjiahan: Get👌 Thank you for the detailed explanation!

packages/core/src/client/hmr.ts
GiveMe-A-Name: image
Web socket connect too slow, so serve side will do first `lazyCompilation` completely quickly. But at this time, web socket is not connected. 
So we need do a hmr action when web socket connect.

.github/workflows/ecosystem-ci.yml
GiveMe-A-Name: Updated

packages/core/src/client/hmr.ts
chenjiahan: Got it, great chart! This is very clear

PR #5788, opened 2025-08-07
feat: add `onBeforeDevCompile` plugin hook

9aoy:
## Summary

`onBeforeDevCompile` is a callback function that is triggered before the dev compile is executed.

```ts
const myPlugin = () => ({
  setup(api) {
    api.onBeforeDevCompile(({ bundlerConfigs }) => {
      console.log('the bundler config is ', bundlerConfigs);
    });
  },
});
```

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #5882, opened 2025-08-19
feat: add new `onAfterDevCompile` plugin hook

chenjiahan:
## Summary

This PR adds a new `onAfterDevCompile` plugin hook to replace the previous `onDevCompileDone` hook.

```js
import type { RsbuildPlugin } from '@rsbuild/core';

const myPlugin: RsbuildPlugin = {
  name: 'my-plugin',
  setup(api) {
    api.onBeforeDevCompile(() => {
      // ...
    });

    // The new hook
    api.onAfterDevCompile(() => {
      // ...
    });

    // The deprecated hook
    api.onDevCompileDone(() => {
      // ...
    });
  },
};
```

The intention of this change is to keep the naming of dev compile related hooks consistent, refer to https://github.com/web-infra-dev/rsbuild/pull/5788#pullrequestreview-3095890071

The previous `onDevCompileDone` hook is still usable. I will explain the relationship between the two in the documentation after v1.5.0 is released.

## Checklist


packages/core/src/hooks.ts
```diff
@@ -500,7 +500,7 @@ export const registerDevHook = ({
     });
 
   const onDone = async (stats: Rspack.Stats | Rspack.MultiStats) => {
-    const p = context.hooks.onDevCompileDone.callBatch({
+    const p = context.hooks.onAfterDevCompile.callBatch({
```
9aoy, line 503: Seems `onDevCompileDone` hook is not triggered anymore?

9aoy (approved):

PR #5943, opened 2025-08-25
feat: add more params for `performance.printFileSize.total`

Timeless0911:
## Summary

Add more params for `performance.printFileSize.total`

- `environmentName`: The unique name of the current environment, used to distinguish and locate the environment
- `distPath`: The output directory relative to the project root

## Checklist

### DIFF
--- a/website/docs/en/config/performance/print-file-size.mdx
+++ b/website/docs/en/config/performance/print-file-size.mdx
@@ -57,7 +57,7 @@ type Total =
   | boolean
   | ((params: {
       environmentName: string;
-      distFolder: string;
+      distPath: string;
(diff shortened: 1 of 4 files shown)


website/docs/en/config/performance/print-file-size.mdx
```diff
@@ -56,6 +56,8 @@ You can customize the output format through the options.
 type Total =
   | boolean
   | ((params: {
+      environmentName: string;
+      distFolder: string;
```
chenjiahan: I prefer to name it `distPath` to align with the `output.distPath` config

Timeless0911: done

PR #5955, opened 2025-08-27
fix: remove default `javascript/auto` type for JavaScript modules

chenjiahan:
## Summary

This PR removes the `type: 'javascript/auto'` from the rule for JavaScript modules.

When Rspack encounters modules with `type: "module"` in package.json or modules with `.mjs` and `.mts` extensions, Rspack will configure these modules with `javascript/esm` by default.

If Rsbuild sets `javascript/auto` for ES modules, this change will break the ESM/CJS interop. For example, when a user writes an `.mjs` file and imports a CJS module, interop runtime should be added. However, with `javascript/auto`, the interop won't be added.

## Related Links

- https://rspack.rs/config/module#ruletype
- https://github.com/fi3ework-reproduction/wrong-external-module-after-multiple-pack

## Checklist


JSerFeng (approved):

chenjiahan:
Closed as it is a potential breaking change (`javascript/esm` and `javascript/auto` behave differently).

PR #6094, opened 2025-09-08
perf: lazy loading memfs dependency

chenjiahan:
## Summary

Lazy loading the `memfs` dependency to reduce the main bundle size and make the dev server faster.

### Before

Screenshot 2025-09-08 at 22 38 38

### After

Screenshot 2025-09-08 at 22 42 46

## Related Links

- https://github.com/web-infra-dev/rsbuild/pull/6092

## Checklist

### DIFF
--- a/packages/core/prebundle.config.mjs
+++ b/packages/core/prebundle.config.mjs
@@ -25,6 +25,7 @@ export default {
     'rspack-manifest-plugin',
     'html-rspack-plugin',
     'mrmime',
+    'memfs',
     'tinyglobby',
     'chokidar',
     'cors',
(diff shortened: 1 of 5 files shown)


TheOnlyArtz:
@chenjiahan Is this feature opt-in by any chance? I suspect it causes us some bugs in production builds at work.

v1.5.4 works perfectly while v1.5.5 throws things like
image

I don't see any other commit which fits this.

EDIT: I went full try hard mode and built rsbuild myself and reverted this commit, didn't seem to help

chenjiahan:
@TheOnlyArtz This PR is not expected to affect the production build, as the memfs package is only required by the Rsbuild dev server. To better understand and debug the issue you’re encountering, it would be very helpful if you could provide a minimal reproduction.

TheOnlyArtz:
> @TheOnlyArtz This PR is not expected to affect the production build, as the memfs package is only required by the Rsbuild dev server. To better understand and debug the issue you’re encountering, it would be very helpful if you could provide a minimal reproduction.

I found out the new tree shaking in rspack caused this :)

PR #6186, opened 2025-09-18
feat: add support for `--config-loader auto`

Timeless0911:
## Summary

Add `auto` for `--config-loader` cli options to use Node.js's native loader to load configuration files first, and fallback to using jiti if it fails.

## Related Links

- RFC: https://github.com/web-infra-dev/rsbuild/discussions/6188

## Checklist

### DIFF
--- a/packages/core/src/loadConfig.ts
+++ b/packages/core/src/loadConfig.ts
@@ -152,10 +152,10 @@ export async function loadConfig({
       if (loader === 'native') {
         logger.error(errorMessage);
         throw err;
-      } else {
-        logger.debug(`${errorMessage}, fallback to jiti.`);
-        logger.debug(err);
       }
+
+      logger.debug(`${errorMessage}, fallback to jiti.`);
+      logger.debug(err);
     }
   }
 


packages/core/src/loadConfig.ts
```diff
@@ -149,10 +158,16 @@ export async function loadConfig({
...
         throw err;
+      } else if (loader === 'auto') {
+        logger.debug(
+          `failed to load file with native loader, fallback to jiti: ${color.dim(configFilePath)}`,
+        );
+        logger.debug(err);
+      } else {
+        logger.debug(
+          `failed to load file with dynamic import: ${color.dim(configFilePath)}`,
+        );
```
chenjiahan: Can we merge the two error logs? For example:

```suggestion
      } else {
        logger.debug(
          `failed to load file with native dynamic import: ${color.dim(configFilePath)}`,
        logger.debug(err);
        );
```

packages/core/src/loadConfig.ts
```diff
@@ -137,8 +138,16 @@ export async function loadConfig({
...
   let configExport: RsbuildConfigExport;
+  let useNative = false;
 
-  if (loader === 'native' || /\.(?:js|mjs|cjs)$/.test(configFilePath)) {
+  // Determine the loading strategy based on the config loader type
+  if (loader === 'native' || loader === 'auto') {
+    useNative = true;
+  } else {
+    useNative = false;
+  }
```
chenjiahan:
```suggestion
  // Determine the loading strategy based on the config loader type
  const useNative = loader === 'native' || loader === 'auto';
```

packages/core/src/loadConfig.ts
Timeless0911: Vibe coding is indeed quite verbose.

packages/core/src/loadConfig.ts
```diff
@@ -138,21 +139,23 @@ export async function loadConfig({
 
   let configExport: RsbuildConfigExport;
 
-  if (loader === 'native' || /\.(?:js|mjs|cjs)$/.test(configFilePath)) {
+  // Determine the loading strategy based on the config loader type
+  const useNative = loader === 'native' || loader === 'auto';
```
9aoy, line 143: Should we detect that the node version has determined whether the current version supports `native`, when the loader is `auto`?

Timeless0911: `process.features.typescript` is enough.

Timeless0911: I will draft another PR to detect this.

PR #6189, opened 2025-09-18
fix: should detect JavaScript runtime when enable `--config-loader auto`

Timeless0911:
## Checklist

### DIFF
--- a/packages/core/src/loadConfig.ts
+++ b/packages/core/src/loadConfig.ts
@@ -140,7 +140,12 @@ export async function loadConfig({
   let configExport: RsbuildConfigExport;
 
   // Determine the loading strategy based on the config loader type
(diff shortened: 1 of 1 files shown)


9aoy (approved):

packages/core/src/loadConfig.ts
```diff
@@ -140,7 +140,12 @@ export async function loadConfig({
   let configExport: RsbuildConfigExport;
 
   // Determine the loading strategy based on the config loader type
-  const useNative = loader === 'native' || loader === 'auto';
+  const useNative =
+    loader === 'native' ||
+    (loader === 'auto' &&
+      (process.features.typescript ||
+        process.versions.bun ||
```
chenjiahan, line 147: We can remove bun: https://github.com/oven-sh/bun/issues/18800

PR #6219, opened 2025-09-22
fix: always output all logs in debug mode

Timeless0911:
## Summary

Always output all logs in debug mode.

## Checklist


e2e/cases/cli/log-level/index.test.ts
```diff
@@ -20,3 +20,18 @@ rspackTest(
     expect(stdout).not.toContain('built in');
   },
 );
+
+rspackTest(
+  'should always verbose logs when debug mode is enabled',
```
chenjiahan:
```suggestion
  'should always print verbose logs when debug mode is enabled',
```

PR #6388, opened 2025-10-20
chore(plugin-vue): pre-copy vue-loader

fi3ework:
## Summary

in Rstest https://github.com/web-infra-dev/rstest/pull/618, introduced a dedicated loader matcher using `with` to implement the sync `importActual` method. however, vue-loader has locked the known properties https://github.com/vuejs/vue-loader/blob/698636508e08f5379a57eaf086b5ff533af8e051/src/pluginWebpack5.ts#L81-L101 so an error will throw when countering `with`.

since vue-loader is not maintained anymore, so pre-copy it in this PR. the pre-copy is just like the `precompile` but without bundling as there are many `require.resovle` usage which will break after bundle.

in this PR, added `with` property to `vue-loader` with pnpm patch, so it will treat `with` as known property.

## Checklist

### DIFF
--- a/packages/plugin-vue/package.json
+++ b/packages/plugin-vue/package.json
@@ -20,7 +20,8 @@
   "main": "./dist/index.cjs",
   "types": "./dist/index.d.ts",
   "files": [
-    "dist"
+    "dist",
+    "compiled"
   ],
   "scripts": {
     "build": "rslib build",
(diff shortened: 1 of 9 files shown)


fi3ework:
closed as prefer using https://github.com/rspack-contrib/rspack-vue-loader.

PR #6464, opened 2025-10-30
fix(ssr): correct inspect offset when loading bundles

9aoy:
## Summary

fix ssr loadBundle inspect offset.

expect inspect `console.log('load SSR')`, but offset:

image

before:
image

after:
image

## Checklist


e2e/cases/server/ssr/rsbuild.config.ts
```diff
@@ -25,6 +26,7 @@ export const serverRender =
 export default defineConfig({
   plugins: [pluginReact()],
   dev: {
+    writeToDisk: inspector.url() !== undefined,
```
chenjiahan, line 30: Could we add some comments

PR #6526, opened 2025-11-07
chore: remove unused hash function config

chenjiahan:
## Summary

Remove unused hash function config, Rspack already uses `xxhash64` by default, see https://github.com/web-infra-dev/rspack/pull/8249

## Checklist


chenjiahan:
webpack still relies on this

PR #6614, opened 2025-11-20
docs: fix JSX import source reference in documentation

htoooth:
## Related Links

Use `importSource` instead of `jsxImportSource`, this is [swc option](https://swc.rs/docs/configuration/compilation#jsctransformreactimportsource).

## Checklist


chenjiahan (approved):
Thank you!

Issue #6100, opened 2025-09-09
[Feature]: Print file size difference since last build


andrzej-early:
### What problem does this feature solve?

`create-react-app` uses `printFileSizesAfterBuild` from `react-dev-utils/FileSizeReporter`. It allows on printing difference in file sizes since last build:

<img width="649" height="150" alt="Image" src="https://github.com/user-attachments/assets/3c4dc713-1502-41c4-94dd-22245f38f004" />

It would be nice to have a similar approach or at least an option to conditionally format the output with some callback in Rsbuild config.

### What does the proposed API look like?

```ts
printFileSize: {
  sizeDifference?: boolean
}
```
or
```ts
printFileSize: {
  customColumns?: Record</* heading */ string, /* cell or row formatter */ ({ filename, size }) => string>
}
```

chenjiahan:
Cool, we can add this feature

lluisemper:
I would like to try and implement this

chenjiahan:
@lluisemper PR welcome!

PR #6635, opened 2025-11-23
feat(core): add showDiff option to printFileSize 

lluisemper:
Add showDiff option to printFileSize for build-to-build size comparison

Add a new `showDiff` option to performance.printFileSize that displays file size changes compared to the previous build. This helps developers track how their changes affect bundle sizes over time.

## Summary

Adds a showDiff option to performance.printFileSize that displays file size changes between builds. Shows color-coded differences (red for increases, green for decreases, "NEW" for new files) in a dedicated "Diff" column. Caches previous build sizes in node_modules/.cache/rsbuild/ for comparison, with automatic hash normalization to match files across builds. Opt-in feature (disabled by default).

E.G.
Screenshot 2025-11-23 at 19 25 32

## Related Links

https://github.com/web-infra-dev/rsbuild/issues/6100

## Checklist

### DIFF
--- a/packages/core/src/plugins/fileSize.ts
+++ b/packages/core/src/plugins/fileSize.ts
@@ -80,6 +80,14 @@ const EXCLUDE_ASSET_REGEX = /\.(?:map|LICENSE\.txt|d\.ts)$/;
 export const excludeAsset = (asset: PrintFileSizeAsset): boolean =>
   EXCLUDE_ASSET_REGEX.test(asset.name);
 
+/** Format a size difference for inline display */
+const formatDiff = (diff: number): string => {
+  const diffStr = calcFileSize(Math.abs(diff));
+  const sign = diff > 0 ? '+' : '-';
+  const colorFn = diff > 0 ? color.red : color.green;
+  return colorFn(`(${sign}${diffStr})`);
+};
+
 const getAssetColor = (size: number) => {
   if (size > 300 * 1000) {
     return color.red;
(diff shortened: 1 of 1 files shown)


packages/core/src/plugins/fileSize.ts
```diff
@@ -18,13 +18,68 @@ import type {
...
   const data = await gzip(input);
   return Buffer.byteLength(data);
 }
 
+/** Get the cache file path for storing previous build sizes */
+function getCacheFilePath(rootPath: string): string {
+  return path.join(
+    rootPath,
+    'node_modules',
+    '.cache',
```
chenjiahan: Use `api.context.cachePath` to replace `rootPath/node_modules/.cache`, see https://rsbuild.rs/api/javascript-api/instance#contextcachepath

packages/core/src/plugins/fileSize.ts
```diff
@@ -119,13 +186,40 @@ async function printFileSizes(
...
+    let diffSize: number | null = null;
+    let diffText: string | null = null;
+    if (showDiff && previousSize !== undefined) {
+      diffSize = size - previousSize;
+      if (diffSize !== 0) {
+        const diffStr = calcFileSize(Math.abs(diffSize));
+        const sign = diffSize > 0 ? '+' : '-';
+        diffText = `${sign}${diffStr}`;
+      } else {
+        diffText = '0 kB';
```
chenjiahan: Can we omit the `0 kB`? It seems like unnecessary information.

chenjiahan (commented):
I feel that create-react-app's logging style is more concise. It does not add an extra “Diff” column but instead shows size changes in parentheses right after the existing size. 

This makes the format more flexible, so we could also show the diff for the gzip size if we want.

Image

lluisemper:
Thank you for reviewing! I have pushed some changes

PR #6712, opened 2025-12-05
feat(core): Add `dev.client.logLevel` option

zalishchuk:
## Summary

Adds an option to control the verbosity of development server browser console logs, similar to the `webpack-dev-server` [`logging`](https://webpack.js.org/configuration/dev-server/#logging) setting.

⚠️ Please include Chinese documentation for `dev.client.logLevel`. I believe that machine-translated documentation won't be as clear as that written by a native speaker. Thanks!

## Related Links
https://webpack.js.org/configuration/dev-server/#logging

## Checklist

### DIFF
--- a/packages/core/src/plugins/fileSize.ts
+++ b/packages/core/src/plugins/fileSize.ts
@@ -7,7 +7,7 @@ import path from 'node:path';
 import { promisify } from 'node:util';
 import zlib from 'node:zlib';
 import { JS_REGEX } from '../constants';
-import { color } from '../helpers';
+import { color, hash } from '../helpers';
 import { getAssetsFromStats, type RsbuildAsset } from '../helpers/stats';
 import { logger } from '../logger';
 import type {
@@ -55,7 +55,10 @@ async function gzipSize(input: Buffer) {
 }
 
 /** Get the cache file path for storing previous build sizes */
-function getSnapshotPath(dir: string): string {
+function getSnapshotPath(dir: string, snapshotHash: string): string {
+  if (snapshotHash) {
+    return path.join(dir, `rsbuild/file-sizes-${snapshotHash}.json`);
+  }
   return path.join(dir, 'rsbuild/file-sizes.json');
 }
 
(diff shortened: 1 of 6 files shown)


packages/core/src/client/log.ts
chenjiahan, line 1: Thank you for your contribution!

I prefer to keep the client code simple and short. Can you reduce unnecessary abstractions to simplify the code?

zalishchuk: Of course, I'll do it by tomorrow. Do you have any particular preferences?

chenjiahan: Here is a minimal logger example. I also avoided using dynamic expressions like `console[method]`:

```js
import type { LogLevel } from '../types';

const LOG_LEVEL: Record<LogLevel, number> = {
  info: 0,
  warn: 1,
  error: 2,
  silent: 3,
};

const prefix = '[rsbuild] ';

export const logger = {
  level: 'info' as LogLevel,
  info(message: string): void {
    if (LOG_LEVEL.info >= LOG_LEVEL[logger.level]) {
      console.log(prefix + message);
    }
  },
  warn(message: string): void {
    if (LOG_LEVEL.warn >= LOG_LEVEL[logger.level]) {
      console.warn(prefix + message);
    }
  },
  error(message: string, err?: unknown): void {
    if (LOG_LEVEL.error >= LOG_LEVEL[logger.level]) {
      console.error(prefix + message, err);
    }
  },
};
```

zalishchuk: Got it, I will simplify the client code a bit later

zalishchuk: Should we get rid of the prefix in logger and just inline it in the message how it was before?

I noticed a small issue with linters warnings/error now having a prefix too, which looks a bit off

Keeps it even simpler, just the log level logic

zalishchuk:
@chenjiahan I've simplified the logger implementation, but we still need to translate the documentation

website/docs/en/config/dev/client.mdx
zalishchuk, line 1: Requires translation.

packages/core/src/client/log.ts
chenjiahan: We should only add a prefix to Rsbuild’s own logs. Adding a prefix to lint errors is unintended behavior and should be considered a bug.

zalishchuk: I reverted to using manual prefixes in [`1a8e6d4` (#6712)](https://github.com/web-infra-dev/rsbuild/pull/6712/commits/1a8e6d4d8bc450c8a7028f0cbd2ded3638c8debe)

There are no prefixes in lint errors or warnings, but the output still follows the current log level, consistent with webpack's behavior

chenjiahan:
> we still need to translate the documentation

Sure! I will add Chinese documentation after merging the PR

chenjiahan (approved):
LGTM, awesome!

PR #6720, opened 2025-12-07
fix(fileSize): Avoid tiny gzip total diffs in file size reports

zalishchuk:
## Summary

The output for file size differences in gzip format included total differences even for very small changes, resulting in noisy logs for insignificant gzip shifts.

The significance check for total gzip size differences has been reused to ensure that only meaningful changes are displayed.

## Checklist


chenjiahan (approved):
Nice catch!

PR #6779, opened 2025-12-15
fix(plugin-vue): treat it as a client build when calling from Rstest

9aoy:
## Summary

fix(plugin-vue): treat it as a client build when calling from rstest, since it always test in dom environment.

reduce `isServerBuild` configuration in user config file.

## Checklist

### DIFF
--- a/packages/plugin-vue/src/index.ts
+++ b/packages/plugin-vue/src/index.ts
@@ -87,7 +87,8 @@ export function pluginVue(options: PluginVueOptions = {}): RsbuildPlugin {
           ...userLoaderOptions.compilerOptions,
         };
         const vueLoaderOptions = {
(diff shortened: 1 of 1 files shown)


packages/plugin-vue/src/index.ts
```diff
@@ -87,6 +87,9 @@ export function pluginVue(options: PluginVueOptions = {}): RsbuildPlugin {
           ...userLoaderOptions.compilerOptions,
         };
         const vueLoaderOptions = {
+          // treat it as a client build when calling from rstest, since it always tests in a DOM environment
```
chenjiahan:
```suggestion
          // Always treat this as a client build when invoked from Rstest,
          // since tests are executed in a DOM-based environment.
```

Issue #6539, opened 2025-11-09
[Bug]: `dev.assetPrefix` leaks to Node env in multi enviroment setup


felixmosh:
### Version

```sh
System:
    OS: macOS 26.0.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 546.80 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 142.0.7444.60
    Firefox: 142.0.1
    Safari: 26.0.1
  npmPackages:
    @rsbuild/core: 1.6.3
```

### Details

when bundling server code, `dev.assetPrefix` is used for node files as well (on dev only) and this breaks imports of worker_threads.

```
import {Worker} from 'worker_threads';

const worker = new Worker(new URL('./worker.ts', import.meta.url));
```

<img width="904" height="321" alt="Image" src="https://github.com/user-attachments/assets/433b7274-ec76-4ffc-a42c-e7249ee4963f" />

<img width="542" height="160" alt="Image" src="https://github.com/user-attachments/assets/3e41a72d-9aa3-47a3-8a80-037b57e7d720" />

### Reproduce link

https://github.com/felixmosh/rsbuild-bug-assetprefix

### Reproduce Steps

On dev
1. pnpm install
2. pnpm dev (files should be generated)
3. exit
4. node dist/server/index.js 

// fails since `dev.assetPrefix` is used (not relevant for node env IMO)

On build
1. pnpm build
3. node dist/server/index.js 

// fails since `output.assetPrefix` on **node** enviroment is ignored and set to be `/` which is an absolute path = new URL uses only it.

chenjiahan:
See: https://github.com/web-infra-dev/rsbuild/issues/3220#issuecomment-2292935595

> This is because node bundles are usually built for SSR and should use the same asset prefix as the web bundles, otherwise there may be some SSR hydration mismatch errors.

chenjiahan:
We might need a solution to allow SSR bundles to correctly reference worker bundles.

PR #6971, opened 2026-01-14
fix(core): allow assetPrefix to be an empty string

xun082:
## Summary

Fixes an issue where `dev.assetPrefix` configuration leaks to Node environment in multi-environment setup, causing Worker imports to fail with "Cannot find module" errors.

**Problem:**
- In development mode, when global `dev.assetPrefix` is set (e.g., `/1.0.0/`), environment-specific `output.assetPrefix` is ignored
- Node environment cannot use relative paths (empty string) for Worker and dynamic imports
- This causes runtime errors: `Error: Cannot find module '/1.0.0/src_server_worker_ts.js'`

**Solution:**
- Environment-specific `output.assetPrefix` now takes priority over global `dev.assetPrefix` in dev mode
- Only applies when explicitly set to non-default values (preserves existing behavior)
- Empty string `assetPrefix` is preserved without formatting to enable relative paths for node targets

**Changes:**
- Modified `getPublicPath()` function in `packages/core/src/plugins/output.ts`
- Added priority logic: environment-specific `output.assetPrefix` (if non-default) → `dev.assetPrefix` → default
- Preserved empty string values without trailing slash formatting

## Related Links

Closes https://github.com/web-infra-dev/rsbuild/issues/6539

## Checklist

- [x] Tests updated (or not required). - All existing unit tests pass (311 tests)
- [x] Documentation updated (or not required). - Code comments added to explain the behavior


xun082:
> I'm concerned this change will break SSR hydration: [#6539 (comment)](https://github.com/web-infra-dev/rsbuild/issues/6539#issuecomment-3514876774)

Thanks for the concern. I've fixed it: when server.base is set, server targets now use it to match the client's public path for SSR hydration, while still preserving empty string for worker_threads when server.base is not set.

chenjiahan (commented):
This PR seems to break down into two parts:

1. Allowing `assetPrefix` to be an empty string. This change makes sense to me, and I’d be happy to see this part merged first.
2. Automatically adjusting `assetPrefix` for the Node target. This part feels a bit too complex, and I find it difficult to clearly explain the behavior in the documentation, so I'm a bit cautious about moving forward with this change.

xun082:
> This PR seems to break down into two parts:
> 
> 1. Allowing `assetPrefix` to be an empty string. This change makes sense to me, and I’d be happy to see this part merged first.
> 2. Automatically adjusting `assetPrefix` for the Node target. This part feels a bit too complex, and I find it difficult to clearly explain the behavior in the documentation, so I'm a bit cautious about moving forward with this change.

Thanks for the feedback! I’ll simplify the changes by keeping the first part (allowing an empty assetPrefix) and removing the automatic adjustment for Node targets. However, there are E2E test failures related to Node target manifest paths — something I haven’t worked with in E2E tests before, so I’m unfamiliar with this error. Do you have any insight into the cause? Should we update the tests to match the new behavior, or do you suggest a different approach for handling manifest paths for Node targets?

chenjiahan:
I will check the failed E2E cases and try to fix them

chenjiahan:
Thank you! I've merged the first part changes

PR #7003, opened 2026-01-20
feat: support server custom message handling by module.hot.on

SyMind:
## Summary

This PR improves the client-side HMR custom message handling by extracting the `module.hot.on` interception logic into a dedicated helper `setupCustomHMRListeners`.

## Checklist

### DIFF
--- a/packages/core/src/client/hmr.ts
+++ b/packages/core/src/client/hmr.ts
@@ -13,51 +13,58 @@ let clearOverlay: undefined | (() => void);
 
 type CustomListenersMap = Map<string, ((data: any) => void)[]>;
 
+type ModuleHot = {
+  on: (event: string, cb: (payload: any) => void) => void;
+  dispose: (cb: () => void) => void;
+};
+
+declare const RSPACK_MODULE_HOT: ModuleHot | undefined;
+
+type InterceptModuleExecutionOptions = {
+  module: {
+    hot: ModuleHot;
+  };
+};
+
+declare const RSPACK_INTERCEPT_MODULE_EXECUTION: ((
+  options: InterceptModuleExecutionOptions,
+) => void)[];
+
 // Install a patched `module.hot.on` that records per-module listeners and
 // removes them from the global map when the module is disposed.
 function setupCustomHMRListeners(customListenersMap: CustomListenersMap): void {
-  // @ts-expect-error
-  RSPACK_INTERCEPT_MODULE_EXECUTION.push(
-    (options: {
-      module: {
-        hot: {
-          on: (event: string, cb: (payload: any) => void) => void;
-          dispose: (cb: () => void) => void;
-        };
-      };
-    }) => {
-      const module = options.module;
+  RSPACK_INTERCEPT_MODULE_EXECUTION.push((options) => {
+    const { module } = options;
 
-      const newListeners: CustomListenersMap = new Map();
+    const newListeners: CustomListenersMap = new Map();
 
-      const addToMap = (
-        map: CustomListenersMap,
-        event: string,
-        cb: (payload: any) => void,
-      ) => {
-        const existing = map.get(event) ?? [];
-        existing.push(cb);
-        map.set(event, existing);
-      };
+    const addToMap = (
+      map: CustomListenersMap,
+      event: string,
+      cb: (payload: any) => void,
+    ) => {
+      const existing = map.get(event) ?? [];
+      existing.push(cb);
+      map.set(event, existing);
+    };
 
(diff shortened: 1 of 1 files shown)


packages/core/rslib.config.ts
```diff
@@ -70,6 +69,53 @@ const pluginFixDtsTypes: rsbuild.RsbuildPlugin = {
...
+                const source = asset.source();
+                if (
+                  source.includes(RSPACK_MODULE_HOT) ||
+                  source.includes(RSPACK_INTERCEPT_MODULE_EXECUTION)
+                ) {
+                  const replacedSource = source
+                    .replaceAll(RSPACK_MODULE_HOT, 'module.hot')
+                    .replaceAll(
+                      RSPACK_INTERCEPT_MODULE_EXECUTION,
+                      RuntimeGlobals.interceptModuleExecution,
```
chenjiahan, line 112: Can we use `define` to inject `RuntimeGlobals.interceptModuleExecution`? It should be simpler.

packages/core/src/client/hmr.ts
```diff
@@ -11,6 +11,51 @@ import { logger } from './log';
 let createOverlay: undefined | ((title: string, content: string) => void);
 let clearOverlay: undefined | (() => void);
 
+type CustomListenersMap = Map<string, ((data: any) => void)[]>;
+
+// Install a patched `module.hot.on` that records per-module listeners and
+// removes them from the global map when the module is disposed.
+function setupCustomHMRListeners(customListenersMap: CustomListenersMap): void {
+  // @ts-expect-error
+  RSPACK_INTERCEPT_MODULE_EXECUTION.push((options: any) => {
```
chenjiahan: Too many `any`, can we use `unknown` instead?

packages/core/src/server/socketServer.ts
```diff
@@ -61,13 +61,21 @@ export type ServerMessageResolvedClientError = {
   };
 };
 
+export type ServerCustomMessage = {
+  type: 'custom';
+  data: {
+    event: string;
+  } & Record<string, any>;
```
chenjiahan: Split `event` and `data` for better extensibility.

```js
data: {
  event: string;
  data?: Record<string, any>;
}
```

packages/core/rslib.config.ts
SyMind: No we can not. Rslib currently doesn't provide a way to preserve `__webpack_require__.*` references in the emitted bundle.

To work around this, we use "magic" placeholder identifiers during authoring (e.g. `RSPACK_MODULE_HOT`, `RSPACK_INTERCEPT_MODULE_EXECUTION`) so rslib won't try to resolve/transform `__webpack_require__.*` at build time.

After bundling, this plugin performs a plain string replacement to swap those placeholders back to the actual Rspack runtime globals.

packages/core/src/client/hmr.ts
```diff
@@ -358,5 +421,10 @@ export function init(
     window.addEventListener('unhandledrejection', onUnhandledRejection);
   }
 
+  // @ts-expect-error
```
chenjiahan: No `@ts-expect-error`, please declare the types

packages/core/src/client/hmr.ts
```diff
@@ -11,6 +11,55 @@ import { logger } from './log';
...
+  RSPACK_INTERCEPT_MODULE_EXECUTION.push(
+    (options: {
+      module: {
+        hot: {
+          on: (event: string, cb: (payload: any) => void) => void;
+          dispose: (cb: () => void) => void;
+        };
+      };
+    }) => {
+      const module = options.module;
```
chenjiahan:
```suggestion
      const { module } = options;
```

e2e/cases/javascript-api/server-custom-message/index.test.ts
```diff
@@ -0,0 +1,70 @@
...
+test('should trigger the client HMR handler when dev server sends a custom message via `sockWrite`', async ({
+  page,
+}) => {
+  const rsbuild = await createRsbuild({
+    cwd: import.meta.dirname,
+  });
+
+  const server = await rsbuild.createDevServer();
+
+  await server.listen();
```
chenjiahan, line 15: Can we just use the `dev` helper?

SyMind: I cannot get dev server instance from `dev()` function.
image

chenjiahan: Feel free to update the helper

chenjiahan:
👍

PR #7547, opened 2026-04-26
feat(source-map): update source map extract schema

zalishchuk:
## Summary

- Add the new flat `output.sourceMap.extract` schema with `type`, `test`, `include`, and `exclude` fields.
- Keep backward compatibility for `extract: true` and the deprecated `extract.js` schema.

## Related Links

- https://github.com/web-infra-dev/rsbuild/pull/7340#issuecomment-4320747930


zalishchuk:
@chenjiahan I've pushed the simplified version (without a `type` version) at https://github.com/web-infra-dev/rsbuild/pull/7547/commits/baf50b989216a213607b592cb456c0e972697649

Currently, adding the `type` isn't necessary since `extractSourceMap` is only relevant for JS. The CSS pipeline automatically loads the original source maps when `sourceMap.css` is enabled, making the `extractSourceMap` rule option unnecessary

Since the shape is flattened, we can always add `type` later if needed. I believe this approach is better than including it now and risking compatibility problems later

Please feel free to revert this commit if you believe we still want to include the `type` option now

packages/core/src/plugins/sourceMap.ts
```diff
@@ -45,44 +52,64 @@ export const pluginSourceMap = (): RsbuildPlugin => ({
...
-        }
-      }
-
-      if (exclude) {
-        for (const condition of exclude) {
-          rule.exclude.add(condition);
-        }
-      }
+      if (include) include.forEach((condition) => rule.include.add(condition));
+      if (exclude) exclude.forEach((condition) => rule.exclude.add(condition));
```
chenjiahan: The prev implementation is more readable:

```suggestion
      if (include) {
        for (const condition of include) {
          rule.include.add(condition);
        }
      }

      if (exclude) {
        for (const condition of exclude) {
          rule.exclude.add(condition);
        }
      }
```

website/docs/en/config/output/source-map.mdx
```diff
@@ -15,6 +15,11 @@ type SourceMap =
       extract?:
         | boolean
         | {
+            /** @default /\.(?:js|mjs|cjs|jsx)$/ */
+            test?: Rspack.RuleSetCondition;
+            include?: Rspack.RuleSetCondition[];
+            exclude?: Rspack.RuleSetCondition[];
+            /** @deprecated Use `test`, `include`, and `exclude` instead. */
```
chenjiahan: I'd prefer to remove the deprecated options from the documentation to keep it clear

website/docs/en/config/output/source-map.mdx
```diff
@@ -182,6 +223,9 @@ export default {
 };
 ```
 
+Use flat `test`, `include`, and `exclude` fields instead.
+:::
+
 ## Version history
```
chenjiahan, line 202: Could you update `Version history` too?

chenjiahan (approved):
LGTM, thank you!

PR #7581, opened 2026-04-30
fix(plugin-svgr): handle string svg base64 encoding

yifancong:
## Summary
  Fix SVG inlining in `plugin-svgr` when loader input is a string.

  This commit updates the encoding logic to handle both input types safely:

  - If `content` is a `Buffer`, keep using `content.toString('base64')`
  - If `content` is a string, use `Buffer.from(content, 'utf8').toString('base64')`

  ## Changes

  - Updated `packages/plugin-svgr/src/assetLoader.ts`
    - Added explicit `Buffer` handling for string vs buffer content before building the data URL.
  - Added `packages/plugin-svgr/tests/assetLoader.test.ts`
    - Verifies correct base64 data URL generation for UTF-8 string SVG input.
    - Verifies correct base64 data URL generation for Buffer SVG input.


chenjiahan (commented):
Could you add an e2e case instead of adding a unit test?

yifancong:
> Could you add an e2e case instead of adding a unit test?

Has added  e2e case for this mr，review again please.

e2e/cases/plugin-svgr/utf8-inline-limit/index.test.ts
```diff
@@ -0,0 +1,16 @@
+import { expect, test } from '@e2e/helper';
+
+test('should not inline utf8 svg when byte length exceeds limit', async ({
```
chenjiahan:
```suggestion
test('should not inline utf8 SVG when byte length exceeds limit', async ({
```

chenjiahan (approved):

chenjiahan:
LGTM 👍

PR #7631, opened 2026-05-10
fix(core): skip compression for 'text/event-stream'

JarvisGG:
Rsbuild currently treats `text/event-stream` as a normal `text/*` response in the default gzip compression logic.

When an SSE endpoint is accessed through the dev server proxy, the upstream response may still be gzipped before being sent back to the browser. For EventSource and other streaming scenarios that depend on timely flushing, this can delay event delivery or cause real-time updates to behave incorrectly.

This PR updates gzipMiddleware to skip gzip compression for text/event-stream responses by default, and adds unit and e2e coverage for the SSE compression behavior.


chenjiahan (approved):
Thank you!

PR #7638, opened 2026-05-12
feat: add output.autoExternal for package dependency externalization

### LINKED ISSUE #5967: [Feature]: Add option to automatically externalize dependencies (similar to Rslib’s autoExternal)

Currently, all dependencies are bundled by default. However, some packages break when they are bundled in Node.js applications. For example, Sentry and OpenTelemetry auto-instrumentation rely on being loaded as external dependencies and will not function correctly if bundled.
To address this, we need an option to automatically externalize dependencies instead of bundling them.

Introduce an option similar to Rslib’s [autoExternal](https://rslib.rs/config/lib/auto-external#libautoexternal), which automatically marks dependencies as external and excludes them from the bundle.
This would allow developers to build Node.js applications without unintentionally breaking packages that require being externalized.

elecmonkey:
## Summary

add `output.autoExternal` for package dependency externalization.

Reverse-port this capability from Rslib back into its upstream project, Rsbuild. In the future, Rslib will consume this capability directly from Rsbuild.

resolve https://github.com/web-infra-dev/rsbuild/issues/5967


elecmonkey:
why cloudflare bot don't come to run for later commits ()

chenjiahan:
> why cloudflare bot don't come to run for later commits ()

The previous comment will be updated for later commits

packages/core/src/helpers/packageJson.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const readPackageJson = (rootPath: string): PackageJson | undefined => {
+  const pkgJsonPath = join(rootPath, 'package.json');
+
+  if (!fs.existsSync(pkgJsonPath)) {
+    return undefined;
+  }
+
+  try {
+    return JSON.parse(fs.readFileSync(pkgJsonPath, 'utf8')) as PackageJson;
```
chenjiahan: Use async fs instead.

fs.readFileSync -> fs.promises.readFile

packages/core/src/helpers/packageJson.ts
```diff
@@ -0,0 +1,25 @@
...
+  dependencies: Record<string, string>;
+  optionalDependencies: Record<string, string>;
+  peerDependencies: Record<string, string>;
+  devDependencies: Record<string, string>;
+}>;
+
+export const readPackageJson = (rootPath: string): PackageJson | undefined => {
+  const pkgJsonPath = join(rootPath, 'package.json');
+
+  if (!fs.existsSync(pkgJsonPath)) {
```
chenjiahan: Use `isFileExists` helper instead

packages/core/src/plugins/externals.ts
```diff
@@ -1,13 +1,129 @@
...
+  return Array.isArray(userExternals)
+    ? [...(userExternals as ExternalItem[]), ...autoExternalRules]
+    : [userExternals as ExternalItem, ...autoExternalRules];
+};
 
 export function pluginExternals(): RsbuildPlugin {
   return {
     name: 'rsbuild:externals',
     setup(api) {
+      const pkgJson = readPackageJson(api.context.rootPath);
```
chenjiahan: Perf tip: should not read package.json if `output.autoExternal` is not enabled

chenjiahan:
@codex

website/docs/zh/config/output/auto-external.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+:::
+
+## 与 Rspack 配置的关系
+
+`output.autoExternal` 不是 Rspack 原生配置项，而是 Rsbuild 基于 Rspack [externals](https://rspack.rs/zh/config/externals) 能力提供的上层封装。开启后，Rsbuild 会读取 `<root>/package.json` 中选中的依赖字段，为每个依赖生成 external 匹配规则，并注入到底层 Rspack `externals` 配置中。
+
+对于每个匹配到的依赖，Rsbuild 会同时生成精确包名匹配和子路径匹配规则。例如 `react` 会匹配 `react`，也会匹配 `react/jsx-runtime` 等子路径导入。
+
+如果你需要完全控制 external 的匹配方式、运行时模块名或与其他 Rspack 配置组合，可以继续使用 [output.externals](/config/output/externals) 或 [tools.rspack](/config/tools/rspack)。
```
chenjiahan: This section is unnecessary and can be removed

packages/core/src/plugins/externals.ts
```diff
@@ -1,13 +1,129 @@
...
+    }, [])
+    .filter((name) => !userExternalKeys.includes(name));
+
+  const uniqueExternals = Array.from(new Set(externals));
+
+  return uniqueExternals.length
+    ? [
+        // Exclude dependencies and subpath imports, e.g. `react`, `react/jsx-runtime`.
+        ...uniqueExternals.map(
+          (dep) => new RegExp(`^${escapeRegExp(dep)}($|\/|\\\\)`),
```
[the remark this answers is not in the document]
elecmonkey: i'm not familiar with regex but they looks like work well ... I don't known why both gemini and copilot are interested in these regex?

website/docs/zh/config/output/auto-external.mdx
```diff
@@ -0,0 +1,134 @@
...
+      react: 'react-18',
+    },
+    autoExternal: true,
+  },
+};
+```
+
+在上面的例子中，`react` 会遵循手动配置的 `output.externals`，不会使用自动生成的 external 规则。
+
+当某个包被配置在 `output.externals` 中时，Rsbuild 不会再为该包生成任何自动 external 规则，包括该包的子路径导入。例如，如果你在 `output.externals` 中配置了 `react`，那么如果你希望 `react/jsx-runtime` 等导入仍然保持 external，也需要将它们添加到 `output.externals` 中。
```
Timeless0911: I think it is not exactly match the code behaviour. For array form like externals: `[{ react: 'React' }]`, `react/jsx-runtime` will still be auto-externalized since the implementation only dedupes exact keys when externals is a plain object. 

The types of externals are quite complex, and I think it would be helpful to tell users to use the command `DEBUG=rsbuild` or `rsbuild inspect` to view the final, effective external configuration if anything does not meet their expects.

elecmonkey: i see... it's only apply to plain obj

```ts
  const userExternalKeys = isPlainObject(userExternals)
    ? Object.keys(userExternals)
    : [];
```

chenjiahan (approved):
LGTM 👍 I will re-check the regexp later

PR #7769, opened 2026-05-27
fix(deps): update @swc/helpers to v0.5.23

CPunisher:
## Summary

`@swc/helpers@0.5.22` contains a decorator helper regression in `_apply_decs_2203_r`. The fixed helper version has been released as `0.5.23`.

I also add a e2e test that will fail with old swc_core (which current rspack depends on) and newer `@swc/helpers`

## Related Links

related: https://github.com/web-infra-dev/rspack/pull/14176


chenjiahan (approved):
LGTM

PR #7776, opened 2026-05-28
test(e2e): migrate runner to rstest

chenjiahan:
## Summary

- Migrates the E2E test runner from `@playwright/test` to Rstest with a standalone `e2e/rstest.config.ts`, while keeping Playwright for browser automation.
- Preserves the existing E2E helper API, including `prepareDist` in test hooks, and removes obsolete Windows skips and migration workarounds.


fi3ework:
there are still some `isolate: false` PRs that haven't been merged, but i have verified the code locally and it runs successfully, achieving an overall performance gain of ~3.5s.

below is the agent's investigation report. i will update this PR once the relevant PRs have landed.

---

## Where the rstest vs Playwright wall-time gap actually is

Same suite (503 files · 907 tests · 14 cores), matched machine load, back-to-back runs. Phases are cut on the same process clock.

| phase | rstest | Playwright | Δ | who wins / why |
|---|---|---|---|---|
| **boot** (start → first test runs) | **1.16s** | **4.69s** | **−3.5s** | **rstest — and this is the entire gap.** Before running the first test, Playwright loads/enumerates every spec file in the main process to collect the test list, so boot scales with file count (7 files → 1.8s, 118 → 2.2s, 504 → 4.7s). rstest rspack-bundles the test code and dispatches to workers, so boot stays flat at ~1.2s. |
| **run** (first test → last test) | 23.45s | 23.79s | ≈0 | **Tie.** Both run the same `rsbuild.build()` calls and the same Playwright browser automation — identical real test work. |
| **drain** (last test → exit) | 0.57s | 0.19s | +0.38s | Playwright slightly faster, negligible. |
| **total** | **25.2s** | **28.6s** | **−3.4s** | equals what boot saves |

Ruled out by experiment — **not** the cause:

- **Parallelism**: rstest defaults to 13 workers, Playwright to 7, but in a 2×2 test both runners gave 7w ≈ 13w (the suite is I/O-bound, 7 workers already saturate). Worker count does not move wall time.
- **babel**: Playwright does use babel, but transforms are content-hash cached on disk — cold vs warm differs by only 0.7s, the wrong order of magnitude.

**TL;DR**: it's not that the runner executes tests faster (the run phase is a tie) — it's that Playwright's startup grows linearly with the number of spec files (it enumerates them all before starting), while rstest's rspack bundle + dispatch keeps boot flat. At 503 files that ~3.5s of startup is the whole difference.

chenjiahan:
Cool 👍

PR #7826, opened 2026-06-03
fix(core): support loadBundle with natural chunk ids

SyMind:
## Summary

- Resolve `loadBundle()` entry chunks by chunk id as well as chunk name.
- Include chunk ids in stats used by `loadBundle()`.
- Add an e2e case covering `loadBundle()` with `optimization.chunkIds: 'natural'` and split chunks.


SyMind:
@codex review

chenjiahan (approved):

PR #7899, opened 2026-06-15
docs: add typescript-go performance tip

Timeless0911:
## Summary

This PR updates the build performance guide to recommend `@rsbuild/plugin-type-check`'s experimental `typescript-go` support when the default TypeScript-based type-checking approach becomes a performance bottleneck. The same guidance is added to both Chinese and English docs.


chenjiahan (approved):
👍

PR #7914, opened 2026-06-16
chore: add plugin npm support metadata

sh962214-hub:
## Summary

- Adds npm support metadata for `@rsbuild/plugin-svelte`
- Adds missing `bugs.url` metadata for `@rsbuild/plugin-solid`

## Why

Both packages already point to this GitHub repository, but their published npm metadata does not expose complete support links. Adding these fields helps npm users find the project site and issue tracker from registry/package manager metadata.

## Validation

- Parsed both package manifests with Node
- Confirmed the expected `homepage` / `bugs.url` values
- Ran `git diff --check`

No runtime code, dependencies, exports, or build outputs were changed.

### DIFF
--- a/packages/plugin-solid/package.json
+++ b/packages/plugin-solid/package.json
@@ -9,6 +9,9 @@
     "url": "https://github.com/web-infra-dev/rsbuild",
     "directory": "packages/plugin-solid"
   },
+  "bugs": {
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
Thank you

PR #7994, opened 2026-06-25
feat(core): support environment scoped exposed APIs

9aoy:
## Summary

This PR adds environment-scoped `expose` / `useExposed` support so integrations such as Rstest can expose per-environment APIs without leaking configuration changes across projects. `useExposed` now prefers an API registered for the current plugin environment and falls back to the global exposed API, while `expose` remains backward-compatible and only scopes APIs when an `environment` option is provided.

## Usage Example

```ts
const rsbuild = await createRsbuild({
  callerName: 'rstest',
  config: {
    environments: {
      project1: {
        plugins: [project1Plugin],
      },
      project2: {
        plugins: [project2Plugin],
      },
    },
  },
});

rsbuild.expose('rstest', project1RstestApi, { environment: 'project1' });
rsbuild.expose('rstest', project2RstestApi, { environment: 'project2' });
```

```ts
const project1Plugin: RsbuildPlugin = {
  name: 'project1-plugin',
  setup(api) {
    const rstestApi = api.useExposed<RstestPluginAPI>('rstest');

    rstestApi?.modifyRstestConfig((config) => {
      config.include.push('**/*.project1.test.ts');
    });
  },
};
```

In this example, `project1Plugin` resolves the `rstest` API exposed for `project1`. It does not receive the API exposed for `project2`, and will fall back to a global exposed API only when no matching environment API exists.

## Related Links

https://github.com/web-infra-dev/rsbuild/blob/main/packages/core/src/pluginManager.ts


9aoy:
@codex review

packages/core/src/initPlugins.ts
```diff
@@ -130,12 +130,48 @@ export function initPluginAPI({
     }
   }) as GetRsbuildConfig;
 
-  const exposed = new Map<string | symbol, any>();
+  type ExposedAPI = {
+    environment?: string;
+    api: unknown;
+  };
+
+  const exposed = new Map<string | symbol, ExposedAPI[]>();
```
chenjiahan: Can we use `${id}-${environmentName}` as the map key?

website/docs/zh/plugins/dev/core.mdx
```diff
@@ -735,9 +746,22 @@ const pluginParent = () => ({
 ```
 
 :::tip
-如果 `api.expose` 被多次调用且使用了相同的 `id`，则后一次调用会覆盖前一次调用暴露的 API。
+如果 `api.expose` 被多次调用且使用了相同的 `id` 和 environment，则后一次调用会覆盖前一次调用暴露的 API。
```
chenjiahan:
```suggestion
如果 `api.expose` 被多次调用且使用了相同的 `id` 和 `environment`，则后一次调用会覆盖前一次调用暴露的 API。
```

website/docs/en/plugins/dev/core.mdx
```diff
@@ -735,9 +746,22 @@ const pluginParent = () => ({
 ```
 
 :::tip
-If `api.expose` is called multiple times with the same `id`, the latter call will overwrite the previously exposed API.
+If `api.expose` is called multiple times with the same `id` and environment, the latter call will overwrite the previously exposed API.
```
chenjiahan:
```suggestion
If `api.expose` is called multiple times with the same `id` and `environment`, the latter call will overwrite the previously exposed API.
```

chenjiahan (approved):

PR #8015, opened 2026-06-30
chore(ci): run lint checks in parallel

chenjiahan:
## Summary

Run the lint, format, and spell checks in a GitHub Actions `parallel` group after the build step so independent lint workflow checks can complete concurrently while preserving separate logs.

## Related Links

https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,11 +37,12 @@ jobs:
       - name: Build
         run: node --run build
 
-      - name: Lint
-        run: node --run lint
+      - parallel:
+          - name: Lint
+            run: node --run lint
 
-      - name: Check Format
-        run: node --run format:check
+          - name: Check Format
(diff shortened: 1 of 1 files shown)


chenjiahan:
Closing this PR because the measured CI data does not show a meaningful improvement.

I compared this PR's `Lint` run with the latest successful pull-request `Lint` runs:

- Current PR: 55s
- Historical median: 56s across 10 recent successful runs
- Historical mean: 55.2s

Although the three checks did start concurrently, the parallel group still took 21s. That is roughly the same as the previous sequential total for `Lint` + `Check Format` + `Check Spell`, because each command became slower when run at the same time. Net result is about 1s faster than the median, which is within normal CI noise and not enough to justify the workflow change.

PR #8023, opened 2026-06-30
chore: relax pnpm engine requirement

chenjiahan:
## Summary

This PR relaxes the root `engines.pnpm` requirement from `>=11.9.0` to `>= 11.0.0` to avoid frequent Codex failures caused by its available pnpm version being lower than the current requirement.

It also disables Renovate updates for the `pnpm` engines field so the relaxed range is not tightened automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   },
   "engines": {
     "node": ">=22.18.0",
-    "pnpm": ">= 11.0.0"
+    "pnpm": ">=11.0.0"
   },
   "packageManager": "pnpm@11.9.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
   },
   "engines": {
     "node": ">=22.18.0",
-    "pnpm": ">=11.9.0"
+    "pnpm": ">= 11.0.0"
```
chenjiahan:
```suggestion
    "pnpm": ">=11.0.0"
```

PR #8027, opened 2026-07-01
test: add dead code elimination e2e case

chenjiahan:
## Summary

Add a minimal e2e regression case that verifies production builds remove unreachable branches guarded by unused exported bindings. The fixture covers both an unused exported mutator and an exported state binding, and checks the generated bundle keeps the reachable markers while dropping the dead-branch markers.

### DIFF
--- a/e2e/cases/optimization/dead-code-elimination/index.test.ts
+++ b/e2e/cases/optimization/dead-code-elimination/index.test.ts
@@ -1,0 +1,9 @@
+import { expect, test } from '@e2e/helper';
+
+test('should remove dead branches guarded by unused exported state', async ({ build }) => {
+  const rsbuild = await build();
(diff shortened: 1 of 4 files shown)


e2e/cases/optimization/dead-code-elimination/index.test.ts
```diff
@@ -0,0 +1,12 @@
+import { expect, test } from '@e2e/helper';
+
+test('should remove dead branches guarded by unused exported bindings', async ({ build }) => {
+  const rsbuild = await build();
+  const indexJs = await rsbuild.getIndexBundle();
+
+  // Keep the assertions marker-based so they do not depend on minified syntax.
```
chenjiahan:
```suggestion
```

PR #8034, opened 2026-07-01
feat(core): support Helium browser

zalishchuk:
## Summary

Add Helium to the list of Chromium browsers that can use Rsbuild's macOS AppleScript open flow.

I noticed this after moving to Helium: `rsbuild --open` opened a new tab each time instead of reusing the existing dev server tab.

This lets Helium users reuse an existing dev server tab when Rsbuild opens the browser through the macOS AppleScript path.

## Related Links

- Helium website: https://helium.computer/


chenjiahan (approved):
Thank you!

PR #8248, opened 2026-08-10
fix(core): restore config key completions in defineConfig

upupming:
## Summary

Since v2.1.8, typing a config key inside `defineConfig({ ... })` stops suggesting config keys and suggests `Function.prototype` members instead.

```ts
export default defineConfig({
  plugins: [pluginReact()],
  to // suggests `apply`, `bind`, `call`, `caller`, `prototype`, `toString`
});
```

The completions are correct on an empty line, and break as soon as the first character is typed. Bisected to #8171, which added const-generic callback overloads before the object overload. While a key is being typed it is not yet a valid `RsbuildConfig` property, so overload resolution falls back to the first matching signature — the callback one — and the contextual type becomes a function.

| `@rsbuild/core` | completions after typing `t` |
| - | - |
| 1.0.19 – 2.1.7 | 17 config keys |
| 2.1.8 – 2.1.10 | 9 `Function.prototype` members |

Simply moving the object overload back to the front reintroduces the widening that #8171 fixed, so this merges the overloads into a single const-generic signature with a conditional return type. That keeps literal inference inside sync and async callbacks, and keeps the existing return types.

## Test Plan

- `e2e/type-tests` keeps the cases from #8171 and adds return type assertions (`RsbuildConfig` / `RsbuildConfigSyncFn` / `RsbuildConfigAsyncFn`) plus a `RsbuildConfigDefinition` value, which covers the overload this PR removes.
- Completions were verified manually through the TypeScript language service, requesting them inside `defineConfig({ ... })` after typing `t` / `to` / `tool`: 9 `Function.prototype` members on the current signature, 17 config keys on this one. Identical results on TypeScript 5.9.3 and 6.0.3, so this is not tied to a compiler version. That check is not automated here — happy to add it if you would like the regression covered.

## Related Links

- #8171


chenjiahan (approved):
Thank you!

PR #8288, opened 2026-08-17
refactor(core): use default loader for dynamic imports

chenjiahan:
## Summary

`CommonJsRunner` currently forwards dynamic `import()` calls from `vm.compileFunction` through a `new Function` wrapper.

This PR uses `vm.constants.USE_MAIN_CONTEXT_DEFAULT_LOADER` so Node.js handles dynamic imports through the main-context default ESM loader, using the compiled filename as the referrer.

## Related Links

- https://nodejs.org/download/release/v20.19.0/docs/api/vm.html#vmconstantsuse_main_context_default_loader
- https://nodejs.org/download/release/v20.19.0/docs/api/vm.html#support-of-dynamic-import-in-compilation-apis
- https://github.com/web-infra-dev/rsbuild/pull/8282

### DIFF
--- a/packages/core/src/server/runner/cjs.ts
+++ b/packages/core/src/server/runner/cjs.ts
@@ -107,17 +107,10 @@ export class CommonJsRunner extends BasicRunner {
       const args = Object.keys(currentModuleScope);
       const argValues = args.map((arg) => currentModuleScope[arg]);
       this.preExecute(file.content, file);
-      const dynamicImport = new Function(
-        'specifier',
-        'return import(specifier)',
-      );
       const fn = vm.compileFunction(file.content, args, {
         filename: file.path,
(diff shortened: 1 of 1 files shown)


chenjiahan:
Closing this PR based on the review feedback.

`vm.constants.USE_MAIN_CONTEXT_DEFAULT_LOADER` is still experimental and emits an `ExperimentalWarning` when a surviving dynamic import is executed. This would add unsolicited stderr output to otherwise successful SSR and server workloads.

The warning-free alternatives either preserve the existing callback's `--experimental-vm-modules` requirement or require custom resolution that could change relative and bare-specifier semantics. Since this standalone change would trade a lint cleanup for a runtime regression, we prefer to keep the current behavior and handle `no-implied-eval` with a narrowly scoped suppression around the intentional fixed-string `Function` bridge.

PR #8290, opened 2026-08-17
fix(core): add transformed ESM fallback for ESM server bundle execution

SyMind:
## Summary

This PR adds a transformed-ESM fallback for evaluating ESM server bundles when Node.js does not expose `vm.SourceTextModule`.

The server runner now selects an execution strategy based on the bundle format and runtime capabilities:

| Bundle output | Runtime capability | Runner |
| --- | --- | --- |
| CommonJS | Any supported Node.js runtime | `CommonJsRunner` |
| ESM | `vm.SourceTextModule` available | `EsmRunner` |
| ESM | `vm.SourceTextModule` unavailable | `TransformedEsmRunner` |

`TransformedEsmRunner` lazily transforms emitted ESM modules into a small async runtime protocol and evaluates them with `AsyncFunction`. Bundle modules are managed by a runner-scoped module graph, while external dependencies are resolved relative to the importer and loaded through native `import()`.

This allows Rsbuild to execute ESM server output without requiring `--experimental-vm-modules`.

## Design

This PR:

- adds `TransformedEsmRunner` as the fallback when `vm.SourceTextModule` is unavailable;
- adds a version-pinned SWC WASM plugin that rewrites ESM imports and exports into the runner protocol;
- preserves live imported and re-exported bindings through namespace property reads;
- preserves unbound function calls and tagged-template calls;
- supports static imports, dynamic imports, cyclic bundle graphs, and top-level `await`;
- provides Node-compatible `import.meta.url`, `import.meta.dirname`, and `import.meta.filename`;
- resolves external modules relative to the importing bundle file;
- validates missing static exports with native ESM-style errors;
- preserves source maps and maps runtime errors back to the original bundle locations;
- builds the WASM plugin in GitHub CI instead of committing the generated binary.

## Why not SystemJS

An earlier implementation used SWC's SystemJS transform and evaluated the resulting `System.register` modules. This works for a closed SystemJS module graph, but it cannot reliably preserve ESM semantics when the graph also contains native external modules.

SystemJS propagates imported values through dependency setters. For a native external module loaded with `import()`, the setter can only be invoked with the module namespace once. If the external module later mutates an exported binding, the native namespace reflects the new value, but the value copied into the SystemJS-generated local binding remains stale.

For example:

```js
// external.mjs
export let ready = false;

export const mark = () => {
  ready = true;
};
```

```js
import { ready, mark } from './external.mjs';

mark();
console.log(ready); // The SystemJS fallback could still observe false.
```

This also affects re-exports and multiple bundle modules importing the same external binding. Native ESM does not expose a subscription mechanism that would let the SystemJS runtime notify dependency setters whenever an external export changes.

The generic SystemJS transform introduced additional complexity:

- Rsbuild had to implement the SystemJS registration, instantiation, setter propagation, cycle handling, and evaluation lifecycle.
- The SWC SystemJS transform required an ES5-target workaround for destructuring assignments.
- Correct execution ordering between bundle modules and native external modules had to be reconstructed by the custom runtime.

The transformed-ESM approach avoids this impedance mismatch. Imported identifiers are rewritten to read directly from the native module namespace, preserving live bindings without requiring external modules to participate in a SystemJS runtime.

### Why not `module.registerHooks()`

A native ESM loader hook could serve bundle output directly from memfs, but it cannot provide safe cache invalidation across rebuilds:

- `module.registerHooks()` was added in Node.js 22.15.0, while `@rsbuild/core` supports Node.js `^20.19.0 || >=22.12.0`.
- Node.js resolves and caches ESM by URL in a cache separate from `require.cache`. Reusing a stable URL such as `/dist/index.js` after a rebuild returns the previous module without calling the `load` hook again.
- Adding a build ID to the URL, such as `/${timestamp}/dist/index.js`, loads the latest source but creates a new cached module graph on every rebuild. Node.js exposes no public API for evicting these ESM cache entries, so old module namespaces and their retained state remain alive and memory usage can grow throughout a long-running watch session.

`SystemJsEvaluator` instead owns a runner-scoped module cache. Replacing the runner releases the whole module graph for garbage collection, while still allowing bundle source to be read from memfs.

References: [Node.js `module.registerHooks()`](https://nodejs.org/download/release/v22.22.0/docs/api/module.html#moduleregisterhooksoptions), [ES modules are cached as URLs](https://nodejs.org/download/release/v22.22.0/docs/api/esm.html#urls), and [ESM uses a separate cache](https://nodejs.org/download/release/v22.22.0/docs/api/esm.html#no-requirecache).


SyMind:
@codex review

SyMind:
@codex review

SyMind:
@codex review

SyMind:
@codex review

chenjiahan (commented):
I’m concerned about two things:

- The potential edge cases introduced by using a plugin to transform ESM into a custom module system
- The various trade-offs of introducing a Wasm plugin, including bundle size, performance, and maintenance costs

PR #8337, opened 2026-08-21
fix(core): use output.target to decide HMR client injection

ramon-villain:
## Summary

- Use Rsbuild's own `output.target` to decide if the dev HMR client should be added to the entries
- Before this, the check only looked for the literal `'web'` in `compiler.options.target`
- If the target is set through `tools.rspack` (for example `target: 'browserslist:...'`), it no longer contains `'web'`, so the HMR client was never added and you had to refresh by hand

## Use case

My app sets the Rspack target through `tools.rspack` to a browserslist query. The HMR runtime was in the bundle, but the HMR client entry was never added, so changes did not show up without a manual refresh.

`output.target` is already normalized to `web`, `node`, or `web-worker`, so it is a safer check than the raw Rspack target.

## Changes

- Replace `isClientCompiler(compiler)` with `config.output.target !== 'web'`
- Remove `isClientCompiler` and its unit test
- Add e2e case `cases/hmr/browserslist-target` that fails on main and passes with this change

## Test plan

- [x] `pnpm e2e cases/hmr/browserslist-target` fails on main, passes with this change
- [x] `pnpm e2e hmr` (22 files, 29 tests)
- [x] core unit tests, lint and type check

### DIFF
--- a/e2e/cases/hmr/browserslist-target/index.test.ts
+++ b/e2e/cases/hmr/browserslist-target/index.test.ts
@@ -1,0 +1,24 @@
+import { expect, test } from '@e2e/helper';
+
+test('should inject HMR client into each entry when target is browserslist', async ({
+  devOnly,
+}) => {
+  const rsbuild = await devOnly({
+    config: {
+      source: {
+        entry: {
+          foo: './src/foo.js',
+          bar: './src/bar.js',
+        },
+      },
+    },
+  });
+
+  const files = rsbuild.getDistFiles();
+  const filenames = Object.keys(files);
+  const fooJs = filenames.find((name) => name.endsWith('foo.js'));
+  const barJs = filenames.find((name) => name.endsWith('bar.js'));
+
+  expect(files[fooJs!].includes('hmr.js')).toBeTruthy();
+  expect(files[barJs!].includes('hmr.js')).toBeTruthy();
+});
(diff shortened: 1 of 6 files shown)


chenjiahan (commented):
Thanks for working on this!

I may be missing some context, but I tested the new e2e cases while keeping the tests unchanged and reverting only the implementation to the existing `EntryPlugin` with `name: undefined`. Both tests still pass, including the actual HMR update test.

In Rspack, `name: undefined` is treated as a global entry dependency, so it is included before every named entry rather than emitted as a separate unused entry. The new named `EntryPlugin` is registered later and appears to append the HMR client instead, which may also affect library entry exports.

Would you mind adding a regression case that fails on the base branch and passes with this change? That would help clarify whether the underlying issue is related to which initial assets are loaded from the manifest.

ramonvillain-st:
@chenjiahan The real problem in my setup was the `target.` I set target: 'browserslist:...' through `tools.rspack`, but `isClientCompiler` only checked for 'web', so the HMR client was never added. The `htmlPlugin: false` part was not the cause.

I changed the PR to a small fix: use Rsbuild's own `output.target` to decide if the HMR client should be added, and removed `isClientCompiler`. The new test `cases/hmr/browserslist-target` fails on main and passes with this change.

chenjiahan (approved):
Thanks for clarifying and updating the PR. I agree this is a real issue that should be fixed.

Using `output.target` addresses the reported Browserslist case, although it may not be precise enough when `tools.rspack` changes the final Rspack target. I’ll merge this PR first, then follow up with a more robust condition based on the resolved compiler platform.

PR #8362, opened 2026-08-24
docs: fix malformed configuration examples

dajiaohuang:
## Summary

Several configuration examples contain invalid JavaScript or do not match the documented Rsbuild option shape, so copying them produces unusable configs. This PR fixes the CSS Modules function example, marks async `bundlerChain` callbacks in both locales, closes the environment-based config object, and corrects the Chinese `distPath` default object.


chenjiahan (approved):
Thank you!

PR #8403, opened 2026-09-01
docs: fix use of deprecated module in rspack.mdx

grossbart:
## Summary

From the Rspack docs: "CircularDependencyRspackPlugin is deprecated. Use CircularCheckRspackPlugin, which has better performance and a more reasonable API design."

## Related links

* https://rspack.rs/plugins/circular-dependency-rspack-plugin (deprecated)
* https://rspack.rs/plugins/circular-check-rspack-plugin


chenjiahan (approved):
Thank you

PR #8412, opened 2026-09-02
chore: enable no-unsafe-call lint rule

chenjiahan:
## Summary

Enable `@typescript-eslint/no-unsafe-call` to prevent unchecked calls on untyped values. Add explicit types at dynamic and third-party boundaries while preserving existing runtime behavior. Migrate the custom-server scripts to TypeScript so their imports are type checked and remain directly runnable with Node.js.

PR #8413, opened 2026-09-02
fix(plugin-svgr): avoid redundant type assertion

SoonIter:
## Summary

Fix the plugin-svgr loader filesystem typing across released and latest Rspack types. Use a local type annotation instead of an assertion so older types remain supported without triggering `no-unnecessary-type-assertion` with the latest `LoaderContext.fs` type.

## Related Links

- https://github.com/web-infra-dev/rsbuild/pull/8412
- https://github.com/web-infra-dev/rspack/pull/15437
- https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/33630889713


chenjiahan (approved):
👍

PR #8462, opened 2026-09-08
docs: correct dev asset prefix resolution

Fyerl:
## Summary

Correct the English and Chinese `dev.assetPrefix` documentation to match how Rsbuild resolves asset URLs:

- Explain that `true` uses the dev server's protocol, hostname, actual listening port, and `server.base`, including the `0.0.0.0` to `localhost` substitution.
- Distinguish an explicit `false` value, which uses `/`, from an omitted value, which inherits `server.base`.
- Correct the HTML examples to use the default `index` entry and preserve a root-relative asset prefix.

## Validation

- Checked the resolution logic in `packages/core/src/plugins/output.ts`, `packages/core/src/defaultConfig.ts`, and `packages/core/src/server/devServer.ts`.
- Ran 11 configuration scenarios against `@rsbuild/core@2.2.4` with Node.js 24.16.0, covering omitted/false/true values, custom base and host, HTTPS, `0.0.0.0`, dynamic ports, root-relative and absolute strings, and the `<port>` placeholder. Confirmed the relevant published implementation matches the current source.
- Compiled a development entry and checked that the generated HTML contains `src="/example/static/js/index.js"`.
- Passed Prettier checks for both changed MDX files with the repository's `singleQuote` setting and `git diff --check`.

- Passed project-configured CSpell checks, MDX compilation, and internal route checks for both changed pages; headings are unchanged.

- The full documentation site build was not run locally; verification fixtures are separate and are not included in the PR.

## Related links

- [dev.assetPrefix](https://rsbuild.rs/config/dev/asset-prefix)
- [server.base](https://rsbuild.rs/config/server/base)

## Checklist

- [x] English and Chinese documentation are synchronized.
- [x] The documented behavior was checked against source and executed configuration/build scenarios.
- [x] This change only updates documentation.

### DIFF
--- a/website/docs/en/config/dev/asset-prefix.mdx
+++ b/website/docs/en/config/dev/asset-prefix.mdx
@@ -34,7 +34,9 @@ export default {
 
 ## Boolean type
 
-If `assetPrefix` is set to `true`, the URL prefix will be `http://localhost:<port>/`:
+If `assetPrefix` is set to `true`, Rsbuild generates an absolute URL prefix from the dev server's protocol, hostname, actual listening port, and [server.base](/config/server/base).
+
+With the default server configuration, the prefix is `http://localhost:<port>/`:
 
 ```ts title="rsbuild.config.ts"
 export default {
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
👍

PR #8463, opened 2026-09-08
docs: correct HTML tag examples and hash behavior

Fyerl:
## Summary

Make the English and Chinese `html.tags` examples match the configuration API and generated HTML.

- Move `output.assetPrefix` to the top level of the CDN example so the public asset receives the documented CDN URL.
- Show custom script tags in `<head>`, their default injection position, and align the handler example's comment with its `d.js` tag.
- Clarify that `hash: true` appends the Rspack compilation hash and that a hash callback returns the complete asset URL.

## Validation

- Built 8 documented configurations across both languages with `@rsbuild/core` 2.2.4 and `@rspack/core` 2.2.3 on Node.js 24.16.0. Checked CDN URLs, tag positions, handler order, and the generated hash against build stats.
- Reproduced the previous CDN example in both languages: `html.output` fails configuration type checking and produces `/service-worker.js` instead of the documented CDN URL.
- Built 2 additional cases confirming that hash callbacks receive the prefixed asset URL and compilation hash and return the final URL.
- TypeScript 5.9.3 checks passed for all 8 documented configurations. Prettier 3.9.6 with the repository's single-quote and heading-case settings and `git diff --check` passed.
- Cross-checked the behavior against the current source. The full documentation site build was not run locally; repository dependencies are not installed. Validation fixtures are separate and are not included in the PR.

- Passed project-configured CSpell checks, MDX compilation, and internal route checks for both changed pages; headings are unchanged.

## Related links

- [HTML tag processing](https://github.com/web-infra-dev/rsbuild/blob/13e1cd4f05f15471e399db01afd2b92886f26119/packages/core/src/rspack-plugins/RsbuildHtmlPlugin.ts)
- [HTML configuration types](https://github.com/web-infra-dev/rsbuild/blob/13e1cd4f05f15471e399db01afd2b92886f26119/packages/core/src/types/config.ts)

## Checklist

- [x] Tests updated (or not required; documentation only).
- [x] Documentation updated (English and Chinese).

### DIFF
--- a/website/docs/en/config/html/tags.mdx
+++ b/website/docs/en/config/html/tags.mdx
@@ -152,9 +152,9 @@ export default {
 Controls whether to add a hash query parameter to asset URLs for cache invalidation, only affects the `src` attribute of the `script` tag and the `href` attribute of the `link` tag.
 
 - `false`: No hash query
-- `true`: Generate hash based on HTML content
+- `true`: Appends the Rspack compilation hash
 - `string`: Uses a custom hash string
-- `function`: Custom hash generation via a function
+- `function`: Returns a URL based on the asset URL and Rspack compilation hash
 
 ```ts title="rsbuild.config.ts"
 export default {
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
👍

PR #8491, opened 2026-09-14
test: split Rsdoctor auto-loading fixtures

yifancong:
## Changes

Split Rsdoctor auto-loading tests into separate fixtures: one with both core and the legacy plugin to verify core takes precedence, and one with only the legacy plugin to verify fallback.

Retain coverage for manually registered plugins and remove the `Module._resolveFilename` mock.

## Validation

- `pnpm build`
- `pnpm e2e cases/profiling/enable-rsdoctor`

External core resolution remains an isolation issue.


chenjiahan (changes requested):
Could we split this into two separate case directories, with one containing both `@rsdoctor/core` and `@rsdoctor/rspack-plugin`, and the other containing only `@rsdoctor/rspack-plugin`?

I’d prefer to express the scenarios through the fixture dependency layout and avoid mocking Node’s private `Module._resolveFilename` API.

chenjiahan (approved):

PR #8493, opened 2026-09-15
test(core): isolate Rsdoctor fallback resolution

yifancong:
## Motivation

The legacy Rsdoctor fixture could resolve an externally injected `@rsdoctor/core` in CI, so it did not reliably exercise the missing-core fallback.

## Changes

Inject the package resolver into the internal Rsdoctor plugin, while keeping the existing Node resolver as the production default. Cover the missing-core fallback with a controlled core unit test and remove the unreliable legacy e2e fixture.

## Validation

Not run locally: the environment is missing the Rspack and Rstack native bindings.


chenjiahan (changes requested):
Could we spy on the existing `require.resolve` from `src/helpers` and test through `createRsbuild()` imported from source? This would avoid adding a resolver parameter solely for testing and manually mocking the plugin API.

9aoy (approved):

chenjiahan (approved):

